PR #34, opened 2023-08-14
Mapped menu items

### LINKED ISSUE #33: Settings Menu Items With Options
**Summary**
_What is the feature all about? Is it a game, a utility, an enhancement, or something else?_

An enhancement to the settings and/or menu interfaces to support settings that aren't just a simple value in a range.

For example, the screensaver setting isn't quite a plain number of seconds, since it should only really be set to a few values within a wide range (e.g. 0, 10, 15, 30, 60, 120, 300) and not in a simple incremental pattern (e.g. 0, 5, 10, 15, 20, ...).

We can add a menu item style that has a list of labels as well as values to give human-readable descriptions to go with actual settings values.

**Technical Spec**
_How will this be implemented?_
_What peripherals will be used, and how?_

In `menu.h` / `menu.c` two new functions would be added to support these new menu item types. The first, to add an item, would behave almost as a combination of the multi-item and the regular setting item, needing both a list of label options and a pointer to the setting bounds, but it would also need a list of actual settings values the same size as the options list.

When the callback is called, the callback will send the selected option as well as the corresponding setting value for that option.

**UI**
_What are the display screens & menus (if applicable)?_
_What user inputs are there for each screen?_

In `menuLogbookRenderer.c`, these new settings would be displayed in the same manner as normal menu items with options.

**Mockups**
_Paste any mockups of screens or anything else here._

**How to Test**
How can we verify the feature is working?

Update the existing screensaver setting to use the mapped settings item implementation and verify that it still allows the screensaver setting to be stored and retrieved.

dylwhich:
### Description

This adds the multi-option settings item menu item as described in #33.

### Test Instructions

Testing:
1. Open the main menu
2. Go to settings
3. Scroll down to **Screensaver** and change the settings by scrolling left or right.
4. Make sure scrolling past the ends doesn't break anything and wraps correctly.
5. Go to the main menu and open some other mode (pong)
6. Go back to the settings menu and check that the screensaver setting is correct.

### Ticket Links

Closes #33 

### Readiness Checklist

### DIFF
--- a/main/menu/menu.c
+++ b/main/menu/menu.c
@@ -379,6 +379,8 @@ void removeSettingsItemFromMenu(menu_t* menu, const char* label)
  * selected label and setting value as the arguments.
  *
  * @param menu The menu to add a settings options item to
+ * @param settingLabel The overall label for this setting. This is what will be
+ *                     passed to the callback when the selected option changes.
  * @param optionLabels All of the labels for each option. The underlying memory
  *                     isn't copied, so these strings must persist for the
  *                     lifetime of the menu.
@@ -390,10 +392,12 @@ void removeSettingsItemFromMenu(menu_t* menu, const char* label)
  * @param currentValue The current value of the setting. Must be one of the values
  *                     in \c optionValues.
  */
-void addSettingsOptionsItemToMenu(menu_t* menu, const char* const* optionLabels, const int32_t* optionValues,
-                                  uint8_t numOptions, const settingParam_t* bounds, int32_t currentValue)
+void addSettingsOptionsItemToMenu(menu_t* menu, const char* settingLabel, const char* const* optionLabels,
+                                  const int32_t* optionValues, uint8_t numOptions, const settingParam_t* bounds,
+                                  int32_t currentValue)
 {
     menuItem_t* newItem  = calloc(1, sizeof(menuItem_t));
+    newItem->label       = settingLabel;
     newItem->options     = optionLabels;
     newItem->settingVals = optionValues;
     newItem->numOptions  = numOptions;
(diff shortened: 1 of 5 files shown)


main/menu/menu.c
```diff
@@ -372,6 +372,94 @@ void removeSettingsItemFromMenu(menu_t* menu, const char* label)
...
+ * each time the setting-options item scrolls or is selected with the newly
+ * selected label and setting value as the arguments.
+ *
+ * @param menu The menu to add a settings options item to
+ * @param optionLabels All of the labels for each option. The underlying memory
+ *                     isn't copied, so these strings must persist for the
+ *                     lifetime of the menu.
+ * @param optionValues The corresponding settings values for each option. The
+ *                     underlying memory isn't copied, so these strings must
+ *                     persist for the lifetime of the menu. These values
```
AEFeinstein, line 387: Missing the end of a sentence?

main/modes/mainMenu/mainMenu.c
```diff
@@ -59,17 +60,48 @@ mainMenu_t* mainMenu;
...
+    30,  // 30sec
+    60,  // 60sec
+    120, // 2min
+    300, // 5min
+};
+
+static const char* const screenSaverSettingsOptions[] = {
+    "Screensaver: Off", "Screensaver: 10s", "Screensaver: 20s", "Screensaver: 30s",
+    "Screensaver: 1m",  "Screensaver: 2m",  "Screensaver: 5m",
+};
```
AEFeinstein, line 82: Hrm, duplicated text. How sane is it to have a single key string, a bunch of value strings, and `strcat()` at runtime?

dylwhich: Ah, yeah that's probably better. And, I think, solves the problem that checking for which label was used is really weird in the callback this way.

AEFeinstein (changes requested):
Looks good except for one small thing. Either the settings shouldn't wraparound when scrolling in the menu (like the non-label ones), or the little arrows should always be drawn. Right now the left and right arrows aren't being drawin at the boundaries, but can still be scrolled in that direction.

AEFeinstein (approved):

PR #39, opened 2023-08-15
Convenience tooling

cnlohr:
### Description

This adds `make usbflash` as a feature to reboot a booted swadge into bootloader, flash it, with `idf.py flash`, reboot back into swadge mode and attach a debug console.  

This currently only runs out-of-box with Linux, though it does work in Windows, not through the makefile interface.  Maybe we should do batch files?


cnlohr:
To be clear - we should discuss more about the right way to handle this on Windows, I feel...

I think a batch file is best, but I don't want to push more root-level files without discussion.

AEFeinstein:
You can alwys put batch files in the `tools` dir to keep the root clean.

cnlohr:
Oooh I like that! I will try that out tonight hopefully.

makefile
```diff
@@ -288,6 +293,7 @@ usbflash :
 	make -C tools/bootload_reboot_stub reboot
 	sleep 2.5
 	make -C tools/swadgeterm monitor
+#endif
```
AEFeinstein: I think you got some C in your makefile, unless this was meant to be commented out?

AEFeinstein: Nevermind, you fixed this. Dunno why github showed me this particular diff.

AEFeinstein:
I had to fidgit with this a bit to get it working on my machine, but I got there (f39d3667a7f96a93562a2b425030cfb8fa3d63a2). I also added documentation, which next time I'll force you to do.

AEFeinstein (approved):

PR #42, opened 2023-08-18
Quick Settings Menu

### LINKED ISSUE #28: Quick Settings Menu
**Summary**
_What is the feature all about? Is it a game, a utility, an enhancement, or something else?_

A settings menu that can be easily called up in the middle of a mode by, for example, holding Start.

This would be a smaller menu than the main menu, maybe even including big icons, that allows the user to quickly change global settings, mainly:
* BGM Mute
* SFX Mute
* LED Brightness
* Screen Brightness
* "Panels Mode" - A quick setting to fully mute and set the LED brightness very low or off

Assuming there's enough room, screensaver and mic gain could also go there.

**Technical Spec**
_How will this be implemented?_
_What peripherals will be used, and how?_

This could be implemented in at least a couple ways:

**A: Completely transparently to the swadge mode:**
Similar to how the "hold select to exit" feature works, the swadge main loop itself would detect the menu being activated. It would work like this:

1. The user holds START, or does something else to activate the quick menu
2. The main loop sets up the quick settings menu
3. The quick settings menu saves the framebuffer into SPIRAM
4. The main loop sets a flag indicating the quick settings menu is active
5. When handling input events, the main loops now sends them to the quick menu instead of the swadge mode
6. When drawing the screen, main loop now calls the quick menu's main loop callback instead of the swadge mode's
7. The quick menu's main loop callback would draw a full-screen settings menu, over top of a dimmed version of the saved framebuffer, sort of like a phone's notification drawer or something
8. When the user exits the quick menu, the flag is cleared
9. The quick settings menu draws the framebuffer back to the screen from SPIRAM
10. The original mode's callbacks are used again instead of the quick menu

Pros:
* Zero effort to add to any swadge mode
* 100% consistent across every mode
* Completely removes need for mode developers to think about global settings or dedicate precious buttons to muting/LED brightness.
* If expanded slightly, this system could be reused to add other "enhancements" to swadge modes without needing to code for it specifically in the swadge mode. An automatic screensaver or background light dances would be another good option for this sort of implementation.

Cons:
* A bit harder to implement
* Might be hard to avoid time jumps affecting the swadge mode still
* Maybe a bit less flexible
* Allowing modes to add their own options to the menu might be a bit more involved

**Cooperatively with the mode**
This would work as an extension the swadge mode would have to actively support, just like the current menus

1. The mode detects that the user wants to open the quick settings menu, however it decides
2. The mode sets a flag to say it's on the quick settings menu
3. When the mode's main loop is called, it checks that flag and calls the quick settings menu main loop callback instead
4. When the mode receives input, it checks the flag and calls the quick settings menu input function instead
5. When the user exits the quick settings menu, the mode clears the flag and resumes as normal.

Pros:
* Less complicated to implement
* Would work just like existing menus
* Would be pretty easy for each mode to add a couple settings of its own
* Maybe adding to each mode individually will decrease the likelihood of the transparent implementation causing bugs with modes that do weird stuff?

Cons:
* Adds a bit more complexity to every swadge mode that uses it
* Would not necessarily be consistent across modes

**UI**
_What are the display screens & menus (if applicable)?_
_What user inputs are there for each screen?_

**Mockups**
_Paste any mockups of screens or anything else here._

**How to Test**
How can we verify the feature is working?

Try it once on each screen in each mode and make sure it works as expected.

dylwhich:
### Description

This is an implementation of the Quick Settings Menu described in #28. It includes a new mode for the quick settings menu, as well as a new menu renderer for the quick setting. Some common menu functionality was also broken out into a separate file to keep things clean and reduce the amount of coupling with the `menuItem_t` struct.

It's not styled at all other than having rounded corners, but it has icons for on/off for LEDs, BGM, and SFX, and more settings and icons can be easily added if needed (TFT backlight?).

Some things are a bit funky and could probably be cleaned up a bit with some API additions:
* Rotating the direction of the menu
* Changing the value of a menu item / its selected option involves a lot of poking at the struct
* The adding extra info to menu items thing that the renderer does is a bit clunky, maybe a data pointer in the menu item?
* Having a separate mute/disable setting for sfx/bgm and leds would make things cleaner and a bit more user friendly

A screenshot of the quick settings menu over top of the pong mode:
the quick settings menu draws over the lazy pong

### Test Instructions

To enter the quick settings menu, hold the START button for .5 seconds from any mode other than the main menu. To exit the quick settings menu, hold the START button for .5 seconds again. A bar should be drawn across the top of the screen to indicate the timer progress. You should not be able to exit to the main menu until the quick settings menu is closed. 

To change between settings, press left and right. To turn a setting on, press up. To turn a setting off, press down. To toggle a setting, press A. If the setting was on (i.e. not set to its min value) at the time the quick settings menu was opened, that value will be used as the "on" value. If the setting was off at the time the quick settings menu was opened, the setting's maximum value will be used as the "on" value.

### Ticket Links

Closes #28 

### Readiness Checklist


AEFeinstein:
I don't think you want to draw progress bars from edge to edge while the menu is open, otherwise it draws over the 'unchanging' background (see top left). You probably want to only draw on the menu itself, and in a different color than the menu background.
image

main/menu/menu_utils.c
```diff
@@ -0,0 +1,132 @@
...
+        buffer[0] = '\0';
+        return buffer;
+    }
+}
+
+/**
+ * @brief Returns whether a menu item is a settings item
+ *
+ * This will return true for items that were added with either ::addSettingsItemToMenu()
+ * or ::addSettingsOptionsItemToMenu()
```
AEFeinstein, line 58: You don't need the `::` for functions with parenthesis. Doxygen sees `()` and does the right thing. You do need `::` for non-function things you want to reference. No harm, no foul though

main/swadge2024.c
```diff
@@ -163,12 +165,19 @@ static swadgeMode_t* cSwadgeMode = &mainMenuMode;
 /// @brief A pending Swadge mode to use after a deep sleep
 static RTC_DATA_ATTR swadgeMode_t* pendingSwadgeMode = NULL;
 
+/// @brief A minimal Swadge mode to be displayed on top of others
+/// Note that it only gets init/exit and mainLoop callbacks.
```
AEFeinstein, line 169: I have a slight preference for using `/** ... */` comments when they're on multiple lines. I'm being stylistically nitpicky, so I must be tired.

main/swadge2024.c
```diff
@@ -163,12 +165,19 @@ static swadgeMode_t* cSwadgeMode = &mainMenuMode;
...
+static swadgeMode_t* overlayMode = NULL;
```
AEFeinstein, line 170: Unless we plan on having multiple different overlays, it'd be a bit cleaner to reference `quickSettingsMode` directly, rather than through this pointer. You'd then need a `bool` which tracks whether the overlay should be shown or not, rather than checking this pointer's null-ness.

main/modes/quickSettings/quickSettings.c
```diff
@@ -0,0 +1,337 @@
...
+// Structs
+//==============================================================================
+
+typedef struct
+{
+    menu_t* menu;                          ///< The menu structure
+    menuQuickSettingsRenderer_t* renderer; ///< Renderer for the menu
+    font_t font;                           ///< The font used for menu text
+
+    wsg_t iconGeneric;
```
AEFeinstein, line 41: What is `iconGeneric` used for?

main/modes/quickSettings/quickSettings.c
```diff
@@ -0,0 +1,337 @@
...
+    // Use the minimum and the current value for the toggles
+    // This way, we get the current value back after toggling off.
+    optionValues[0] = bounds->min;
+    optionValues[1] = currentValue;
+
+    // However, if it was already off / at the minimum, we need it to be something, so just use the max
+    if (currentValue <= bounds->min)
+    {
+        optionValues[1] = bounds->max;
+    }
```
AEFeinstein, line 137: Yeah, this is where storing the quick setting binary value separately from the actual setting would be nice. Or maybe the quick setting should have the same range of values for each setting as the main settings menu? Binary on/off is quicker, but you get more flexibility with more options, and you can use the current NVS setting as-is.

AEFeinstein:
I'll second adding TFT backlight settings!

main/modes/quickSettings/quickSettings.c
```diff
@@ -0,0 +1,337 @@
...
+    loadWsg("ledsEnabled.wsg", &quickSettings->iconLedsOn, true);
+    loadWsg("ledsDisabled.wsg", &quickSettings->iconLedsOff, true);
+    loadWsg("musicEnabled.wsg", &quickSettings->iconBgmOn, true);
+    loadWsg("musicDisabled.wsg", &quickSettings->iconBgmOff, true);
+    loadWsg("sfxEnabled.wsg", &quickSettings->iconSfxOn, true);
+    loadWsg("sfxDisabled.wsg", &quickSettings->iconSfxOff, true);
+
+    // Initialize the menu
+    quickSettings->menu     = initMenu(quickSettingsName, quickSettingsMenuCb);
+    quickSettings->renderer = initMenuQuickSettingsRenderer(&quickSettings->font);
```
AEFeinstein, line 167: I'm so happy that there is a second renderer. It really validates splitting the data structure from the UI 😊

main/modes/quickSettings/quickSettings.c
```diff
@@ -0,0 +1,337 @@
...
+                evt.button = PB_LEFT;
+                break;
+
+            case PB_LEFT:
+                evt.button = PB_UP;
+                break;
+
+            case PB_RIGHT:
+                evt.button = PB_DOWN;
+                break;
```
AEFeinstein, line 278: You sly devil!

main/modes/quickSettings/menuQuickSettingsRenderer.c
```diff
@@ -0,0 +1,166 @@
...
+
+/// The height of the panel, from the top of the screen
+#define PANEL_H 60
+
+/// The panel corner-radius
+#define PANEL_R 15
+
+#define TEXT_MARGIN 10
+
+#define PANEL_BG_COLOR   c333
```
AEFeinstein, line 32: This color shouldn't match the button-held-bar color in `swadge2024.c`

main/modes/quickSettings/menuQuickSettingsRenderer.c
```diff
@@ -0,0 +1,166 @@
...
+//==============================================================================
+
+typedef struct
+{
+    /// @brief The menu item label these
+    const char* label;
+    /// @brief The icon to draw when the setting is on
+    const wsg_t* onWsg;
+    /// @brief The icon to draw when the setting is off
+    const wsg_t* offWsg;
```
AEFeinstein, line 51: If you do like the option of having the full range of settings, you can still use the on/off images, and have the off image only apply to the min setting.

main/modes/quickSettings/menuQuickSettingsRenderer.c
```diff
@@ -0,0 +1,166 @@
...
+{
+    menuQuickSettingsRenderer_t* renderer = calloc(1, sizeof(menuQuickSettingsRenderer_t));
+
+    renderer->font = font;
+    return renderer;
+}
+
+void deinitMenuQuickSettingsRenderer(menuQuickSettingsRenderer_t* renderer)
+{
+    clear(&renderer->iconMap);
```
AEFeinstein, line 98: I think this is a leak. `clear()` doesn't `free()` values in the linked list, it only frees the nodes. There's no guarantee the pointers `push()`'d to the list were dynamically allocated.

see the `callco()`:
```c
void quickSettingsRendererAddIcon(menuQuickSettingsRenderer_t* renderer, const char* label, const wsg_t* onWsg,
                                  const wsg_t* offWsg)
{
    quickSettingsItemInfo_t* info = calloc(1, sizeof(quickSettingsItemInfo_t));

    info->label  = label;
    info->onWsg  = onWsg;
    info->offWsg = offWsg;

    push(&renderer->iconMap, info);
}
```

AEFeinstein (changes requested):

main/modes/quickSettings/menuQuickSettingsRenderer.c
AEFeinstein: Yeah, confirmed by `drmemory.exe`. I wonder why `libasan` didn't whine about it?
```
Error #155: LEAK 24 direct bytes 0x000001f7907da7b0-0x000001f7907da7c8 + 0 indirect bytes
# 0 replace_calloc                             [d:\a\drmemory\drmemory\common\alloc_replace.c:2620]
# 1 quickSettingsRendererAddIcon               [main/modes/quickSettings/menuQuickSettingsRenderer.c:159]
# 2 quickSettingsEnterMode                     [main/modes/quickSettings/quickSettings.c:187]
# 3 app_main                                   [main/swadge2024.c:423]
# 4 main                                       [emulator/src/emu_main.c:104]

Error #156: LEAK 24 direct bytes 0x000001f7907da830-0x000001f7907da848 + 0 indirect bytes
# 0 replace_calloc                             [d:\a\drmemory\drmemory\common\alloc_replace.c:2620]
# 1 quickSettingsRendererAddIcon               [main/modes/quickSettings/menuQuickSettingsRenderer.c:159]
# 2 quickSettingsEnterMode                     [main/modes/quickSettings/quickSettings.c:189]
# 3 app_main                                   [main/swadge2024.c:423]
# 4 main                                       [emulator/src/emu_main.c:104]

Error #157: LEAK 24 direct bytes 0x000001f7907da8b0-0x000001f7907da8c8 + 0 indirect bytes
# 0 replace_calloc                             [d:\a\drmemory\drmemory\common\alloc_replace.c:2620]
# 1 quickSettingsRendererAddIcon               [main/modes/quickSettings/menuQuickSettingsRenderer.c:159]
# 2 app_main                                   [main/swadge2024.c:423]
# 3 main                                       [emulator/src/emu_main.c:104]
```

main/swadge2024.c
dylwhich: Ah yeah, that does seem a bit silly doesn't it

main/modes/quickSettings/quickSettings.c
dylwhich: I wanted to have a default case and I was at one point going to add a few more settings and didn't want to come up with icons for those too. I can get rid of it if it doesn't end up being used.

main/modes/quickSettings/quickSettings.c
dylwhich: Yeah, I think I'll do it like that. It is a bit weird just being able to toggle it between two values and there's no reason it shouldn't be able to do that.

main/modes/quickSettings/quickSettings.c
AEFeinstein: Ah, that makes sense. I'm always on the lookout for unused assets before bloat becomes a problem (just ask @Brycey92)

main/modes/quickSettings/menuQuickSettingsRenderer.c
dylwhich: This has been done now! And I think it's much better like this

main/modes/quickSettings/menuQuickSettingsRenderer.c
dylwhich: Should be fixed now! Good catch

main/modes/quickSettings/quickSettings.c
dylwhich: What can I say, I'm a sucker for a well defined interface!

dylwhich:
Now with backlight and styled to match the logbook menu:

image

main/modes/quickSettings/quickSettings.c
```diff
@@ -68,23 +80,19 @@ static int32_t quickSettingsMenuFlipItem(const char* label);
...
-    quickSettingsOn,
-};
+static const char quickSettingsLedsOff[] = "LEDs Off";
+static const char quickSettingsLedsMax[] = "LED Brightness: Max";
+static const char quickSettingsSfxMuted[] = "SFX Muted";
+static const char quickSettingsSfxMax[]= "SFX Volume: Max";
+static const char quickSettingsBgmMuted[] = "Music Muted";
+static const char quickSettingsBgmMax[]= "Music Volume: Max";
+static const char quickSettingsBacklightOff[] = "Screen Backlight Off";
+static const char quickSettingsBacklightMax[] = "Screen Brightness: Max";
```
AEFeinstein: It would be real nice to have consistent, shared strings between main menu settings and quick settings.

main/swadge2024.c
```diff
@@ -402,40 +401,31 @@ void app_main(void)
...
 
                     // Reset the count
                     timePausePressed = 0;
                 }
                 else
                 {
-                    int16_t numPx = (tHeldUs * TFT_WIDTH) / PAUSE_TIME_US;
-                    fillDisplayArea(0, 0, numPx, 10, c333);
+                    int16_t numPx = (tHeldUs * (TFT_WIDTH - 60)) / PAUSE_TIME_US;
+                    fillDisplayArea(30, 0, 30 + numPx, 10, c333);
```
AEFeinstein, line 425: You should `#define` these numbers in a header to ensure the width of the bar matches the width of the quick settings menu.

AEFeinstein:
Oh man that stylized border is _hot_

main/modes/quickSettings/quickSettings.c
dylwhich: Ah, yeah... I went for longer strings here since the main menu's settings are very abbreviated and there's not a ton of room for longer names with the logbook font. I'll see what I can come up with that's consistent and fits, maybe I can use the logbook font here too then.

AEFeinstein: OK, you thought this through already. Smaller font does mean more space to work with, so that's fine. You don't need to bring them in line.

main/swadge2024.c
AEFeinstein: It is consistent at least. You could mathematically figure out that `30` too.
```c
/// The width of the panel, centered in the screen
#define PANEL_W (TFT_WIDTH - 60)
```

AEFeinstein: or do
```c
#define PANEL_MARGIN 30
#define PANEL_WIDTH (TFT_WIDTH - (2 * PANEL_MARGIN))
```

main/modes/quickSettings/quickSettings.c
```diff
@@ -0,0 +1,437 @@
...
+    // Use SPI because we're not the only mode, I guess?
+    loadWsg("defaultSetting.wsg", &quickSettings->iconGeneric, true);
+    loadWsg("ledsEnabled.wsg", &quickSettings->iconLedsOn, true);
+    loadWsg("ledsDisabled.wsg", &quickSettings->iconLedsOff, true);
+    loadWsg("musicEnabled.wsg", &quickSettings->iconBgmOn, true);
+    loadWsg("musicDisabled.wsg", &quickSettings->iconBgmOff, true);
+    loadWsg("sfxEnabled.wsg", &quickSettings->iconSfxOn, true);
+    loadWsg("sfxDisabled.wsg", &quickSettings->iconSfxOff, true);
+    loadWsg("backlightEnabled.wsg", &quickSettings->iconTftOn, true);
+    loadWsg("backlightDisabled.wsg", &quickSettings->iconTftOff, true);
```
AEFeinstein, line 177: These are leaking! `libasan` caught it. Make sure to peek at your terminal after closing.

AEFeinstein (changes requested):

main/modes/quickSettings/quickSettings.c
dylwhich: Ah, you know, I think asan doesn't actually print anything if you kill the emulator with Ctrl+C, on linux at least...

AEFeinstein: TIL

dylwhich:
Slightly updated look: 
image

And rounded the edges of the progress bar, because it starts in the middle of the screen instead of being obscured by the rounded corners: 
image
image

AEFeinstein (approved):

PR #44, opened 2023-08-23
Stereo bzr

### LINKED ISSUE #10: Improve Buzzer Performance
* Ideally, would be able to change the frequency every 1ms
    * Being able to change frequency rapidly would allow for percussive sounds (kick drums and white noise)
* LEDC max frequency is 40MHz with fixed duty cycle
    * if we aim for 1MHz max frequency, we can get 6-bit duty cycle resolution
        * 44100KHz would be 10-bit
    * (the equation is duty_cycle_resolution = floor(log2(80MHz / max_freq)) )
* Using the ULP coprocessor, it may be possible to handle all the audio PWM instead of using ledc which may take longer to update
    * not testable as-is since GPIO0-21 are available and buzzer is on GPIO40, but SAO is available to use on GPIO17 and 18

AEFeinstein:
### Description

 * Second buzzer was enabled
 * SNG file format, processor, and loader were all updated for multiple tracks
 * Update emulator with stereo sound (it works but sounds bad)
 * Documentation was updated
 * Allowed TFT backlight to use a specific ledc_timer_t
 * Fixed `.gitignore` for spiffs_file_preprocessor
 * Fixed USB formatting
 * Fixed spelling

### Test Instructions

Play the Pong mode, which uses a stereo song. Cover each buzzer with your thumb to confirm both are playing different tunes.

### Ticket Links

Closes #10

### Readiness Checklist


components/hdw-bzr/hdw-bzr.c
```diff
@@ -13,56 +15,82 @@
 // Defines
 //==============================================================================
 
-// For LEDC
 #define LEDC_MODE LEDC_LOW_SPEED_MODE //!< Low speed mode is sufficient
 
 //==============================================================================
-// Enums
+// Structs
```
dylwhich, line 21: That's no struct!

main/modes/mainMenu/mainMenu.c
```diff
@@ -219,12 +219,12 @@ static void mainMenuCb(const char* label, bool selected, uint32_t settingVal)
         else if (bgmVolSettingLabel == label)
         {
             setBgmVolumeSetting(settingVal);
-            bzrPlayBgm(&mainMenu->jingle);
+            bzrPlaySfx(&mainMenu->jingle, BZR_LEFT);
```
dylwhich, line 222: Shouldn't this still be `bzrPlayBgm`?

components/hdw-bzr/hdw-bzr.c
```diff
@@ -13,56 +15,82 @@
...
 {
-    const song_t* song;  ///< The song_t being played
-    uint32_t note_index; ///< The index of the current musicalNote_t in the song
-    int64_t start_time;  ///< The time the current musicalNote_t started in the song
-    uint16_t volume;     ///< This track's volume, from 0 (off) to 4096 (max)
-} buzzerTrack_t;
+    NO_LOOP,
+    MONO_LOOP,
+    STEREO_LOOP
+} loopType_t;
```
dylwhich, line 32: What do these actually mean? Is it to keep track of whether both tracks loop together or separately?

dylwhich (approved):

dylwhich:
Tested it with the emulator since I obviously lack hardware. It works great with headphones though!

components/hdw-bzr/hdw-bzr.c
AEFeinstein: D'oh, will fix

components/hdw-bzr/hdw-bzr.c
AEFeinstein: These should have been (and will be) removed. There was a moment where even though the time lengths of L and R tracks were identical, they would loop at slightly different times and desync. I put in a hack so that both tracks would loop whenever either finished.

It turned out the real problem was that `start_time` was being set to `cTime` rather than accumulating `noteTimeUs` (see a4a170d). This would lead to small drifts because notes are checked every 5ms, but note lengths do not need to be multiples of 5ms. Now the two tracks keep in sync, and the song is played at _exactly_ the right speed.

main/modes/mainMenu/mainMenu.c
AEFeinstein: Yeah

components/hdw-bzr/hdw-bzr.c
dylwhich: Ahhh okay, that makes sense!

main/modes/mainMenu/mainMenu.c
AEFeinstein: And it should be `BZR_STEREO`. I was only using L/R for testing here

AEFeinstein:
It sounded good with headphones? It sounds pretty junky on windows and through WSL, though I'm surprised WSL sound works at all.

dylwhich:
> It sounded good with headphones? It sounds pretty junky on windows and through WSL, though I'm surprised WSL sound works at all.

Yeah, it seems to work pretty well with pulseaudio. It has small glitches here and there but it sounds about as good as you could expect for square waves.

AEFeinstein:
Fascinating. I may need to bother @cnlohr about crummy windows sound sometime.

PR #51, opened 2023-08-27
Add convenience clean/export/install/flash scripts

Brycey92:
### Description

Add convenience clean/export/install/flash scripts

### Test Instructions

Tested all scripts in their respective shells.

### Readiness Checklist


tools/export.ps1
```diff
@@ -0,0 +1,13 @@
...
+
+pushd ~/esp/esp-idf
+
+# Install tools
+./install.ps1
+
+# Export paths and variables
+./export.ps1
+
+popd
```
AEFeinstein, line 13: I don't get what this is doing. What's the point of making an `~/esp/` folder and calling a script that may not exist in an empty folder?

tools/clean.bat
```diff
@@ -0,0 +1,5 @@
+make clean
+idf.py clean fullclean
+rmdir /s /q build
+rmdir /s /q spiffs_image
+mkdir spiffs_image
```
AEFeinstein, line 5: This should be a `makefile` target, like `make fullclean`, not a batch file. That way it's cross-platform. This is what you want:

```
fullclean: clean
	idf.py fullclean
	git clean -fX
	-@rm -rf ./docs/html
	-@rm -rf ./spiffs_image/*
	$(MAKE) -C ./tools/sandbox_test clean
	$(MAKE) -C ./tools/hidapi_test clean
	$(MAKE) -C ./tools/bootload_reboot_stub clean
	$(MAKE) -C ./tools/font_maker clean
	$(MAKE) -C ./tools/swadgeterm clean
	$(MAKE) -C ./tools/reboot_into_bootloader clean
	$(MAKE) -C ./tools/spiffs_file_preprocessor clean
```

AEFeinstein:
These scripts may work on your machine, but they're all dependent on an installation in `%UserProfile%\esp\esp-idf\`, which isn't guaranteed.

I'm also not a fan of the one-line scripts (`install.bat`, `export.bat`) that are basically symlinks. Adding `%UserProfile%\esp\esp-idf\` to your path is the cleaner solution. That allows you to call the IDF scripts from anywhere.

`flash.bat` is invalidated by `make usbflash` (try it out!)

tools/export.ps1
Brycey92: I think I must have stripped down the dev env setup script from last year to make this. Not sure why it does half the stuff it does, so I'll delete it.

AEFeinstein (approved):

PR #72, opened 2023-09-16
Breakout - initial PR

JVeg199X:
Not included: ability to change framerate to 60FPS


main/modes/breakout/aabb_utils.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include <stdint.h>
+#include <stdbool.h>
+#include <palette.h>
+
+typedef struct
+{
+    int32_t x;
+    int32_t y;
+} vector_t;
```
AEFeinstein, line 12: We do have `vector2d.h` and `geometry.h` but it seems that we all like re-implementing these things anyway.

main/modes/breakout/common_typedef.h
AEFeinstein, line 1: This is fine for now, but a generically named file like this has a chance of colliding with another mode. `breakout_typedef.h` is more unique.

AEFeinstein (approved):

PR #84, opened 2023-09-25
Fix quick settings

AEFeinstein:
### Description

Fixed issues relating to the quick settings menu and the buzzer including
* Added ability to pause buzzer
    * Pause and resume when showing and hiding the quick settings menu
* Ensure quick settings are not swapped in the middle of a mode's main loop
* Added select button override

### Test Instructions

Run on real hardware, enter and exit quick settings while music is playing (Pong mode does it well)

### Ticket Links

n/a

### Readiness Checklist


components/hdw-bzr/hdw-bzr.c
```diff
@@ -402,28 +409,44 @@ void IRAM_ATTR bzrStopNote(buzzerPlayTrack_t track)
 static bool IRAM_ATTR buzzer_check_next_note_isr(gptimer_handle_t timer, const gptimer_alarm_event_data_t* edata,
                                                  void* user_ctx)
 {
-    // Don't do much if muted
+    // Don't do much if muted or paused
     if ((0 == bgmVolume) && (0 == sfxVolume))
```
dylwhich, line 413: Should this be checking `bzrPaused` too? Not sure if that's what the comment is saying since there is a check down below.

AEFeinstein: Oh, I was initially checking `bzrPaused` here, but if that happens, `tLastLoopUs` doesn't get updated so when you unpause there's a massive time skip which sounds... not good on the buzzer. I'll fix the comment

AEFeinstein: I thought about this for 30s more and moved the muted check to where the pause check is. No giant time skips for anyone!

JVeg199X (approved):
Looks good, and works from what I've been able to test so far.
Also, being able to pause songs is great!

PR #95, opened 2023-10-17
Fix settings gripes

dylwhich:
### Description

Fixes a couple settings-related issues I noticed while working on real hardware, and adds a bit more polish to quick and regular settings.

Fixed issues:
* TFT and LED brightness settings didn't actually apply until you changed them after boot, since they initialize after reading the settings and would only use defaults.
* BGM was played with SFX volume on hardware

Enhancements:
* Quick settings menu now does an LED pattern while setting LED brightness
* Added feature to save and restore the buzzer state, so quick settings doesn't reset the underlying mode's music
* Quick settings menu plays sound while changing volume (does not work, at all -- seems to be an issue with quick settings specifically, or I am doing something dumb)
* Main menu settings only play sounds when levels are changed, not just scrolling past them

### Test Instructions

On the real swadge, change the TFT and LED brightness settings however you want, then reboot the swadge. You should see that the settings are set properly from the beginning, instead of using the defaults until you change them.

Go into the main menu settings and scroll past the SFX/BGM settings. Note that they don't play sound until you actually change the value.

### Readiness Checklist

### DIFF
--- a/components/hdw-bzr/hdw-bzr.c
+++ b/components/hdw-bzr/hdw-bzr.c
@@ -567,3 +567,42 @@ void bzrResume(void)
         }
     }
 }
+
+/**
+ * @brief Save the state of the buzzer so that it can be restored later, perhaps
+ * after playing a different sound.
+ *
+ * @return A void-pointer which can be passed back to bzrRestore()
+ */
+void* bzrSave(void)
+{
+    bzrPause();
+
+    bzrTrack_t* result = malloc(sizeof(bzrTrack_t) * NUM_BUZZERS * 2);
+    for (uint16_t bIdx = 0; bIdx < NUM_BUZZERS; bIdx++)
+    {
+        memcpy(result + bIdx * 2, &buzzers[bIdx].bgm, sizeof(bzrTrack_t));
+        memcpy(result + bIdx * 2 + 1, &buzzers[bIdx].sfx, sizeof(bzrTrack_t));
+    }
+
+    return (void*)result;
+}
+
+/**
+ * @brief Restore the state of the buzzer from a void-pointer returned by bzrSave()
+ *
+ * The data passed pointer will be freed by this call.
+ *
+ * @param data The saved state of the buzzer, returned by bzrSave()
+ */
+void bzrRestore(void* data)
+{
+    bzrTrack_t* buzzerState = (bzrTrack_t*)data;
+    for (uint16_t bIdx = 0; bIdx < NUM_BUZZERS; bIdx++)
+    {
+        memcpy(&buzzers[bIdx].bgm, buzzerState + bIdx * 2, sizeof(bzrTrack_t));
+        memcpy(&buzzers[bIdx].sfx, buzzerState + bIdx * 2 + 1, sizeof(bzrTrack_t));
+    }
+
+    free(data);
+}
--- a/components/hdw-bzr/hdw-bzr.c
+++ b/components/hdw-bzr/hdw-bzr.c
@@ -433,7 +433,7 @@ static bool IRAM_ATTR buzzer_check_next_note_isr(gptimer_handle_t timer, const g
             // Try playing SFX first
             bool sfxIsActive = buzzer_track_check_next_note(&bzr->sfx, bIdx, sfxVolume, true, tElapsedUs);
             // Then play BGM if SFX isn't active
-            bool bgmIsActive = buzzer_track_check_next_note(&bzr->bgm, bIdx, sfxVolume, !sfxIsActive, tElapsedUs);
+            bool bgmIsActive = buzzer_track_check_next_note(&bzr->bgm, bIdx, bgmVolume, !sfxIsActive, tElapsedUs);
 
             // If nothing is playing, but there is BGM (i.e. SFX finished)
             if ((false == sfxIsActive) && (false == bgmIsActive) && (NULL != bzr->bgm.sTrack))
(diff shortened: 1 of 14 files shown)


components/hdw-bzr/hdw-bzr.c
```diff
@@ -433,7 +433,7 @@ static bool IRAM_ATTR buzzer_check_next_note_isr(gptimer_handle_t timer, const g
             // Try playing SFX first
             bool sfxIsActive = buzzer_track_check_next_note(&bzr->sfx, bIdx, sfxVolume, true, tElapsedUs);
             // Then play BGM if SFX isn't active
-            bool bgmIsActive = buzzer_track_check_next_note(&bzr->bgm, bIdx, sfxVolume, !sfxIsActive, tElapsedUs);
+            bool bgmIsActive = buzzer_track_check_next_note(&bzr->bgm, bIdx, bgmVolume, !sfxIsActive, tElapsedUs);
```
AEFeinstein, line 436: Good catch

components/hdw-bzr/hdw-bzr.c
```diff
@@ -567,3 +567,42 @@ void bzrResume(void)
...
     }
 }
+
+/**
+ * @brief Save the state of the buzzer so that it can be restored later, perhaps
+ * after playing a different sound.
+ *
+ * @return A void-pointer which can be passed back to bzrRestore()
+ */
+void* bzrSave(void)
```
AEFeinstein, line 577: Ah, not quite the same as `bzrPause()` and `bzrResume()`. Gotcha

components/hdw-bzr/hdw-bzr.c
```diff
@@ -567,3 +567,42 @@ void bzrResume(void)
...
+{
+    bzrPause();
+
+    bzrTrack_t* result = malloc(sizeof(bzrTrack_t) * NUM_BUZZERS * 2);
+    for (uint16_t bIdx = 0; bIdx < NUM_BUZZERS; bIdx++)
+    {
+        memcpy(result + bIdx * 2, &buzzers[bIdx].bgm, sizeof(bzrTrack_t));
```
AEFeinstein, line 584: I'm not the biggest fan of pointer arithmetic. `&result[bIdx * 2]` reads clearer to me. Same goes for the line below and in `bzrRestore()`.

AEFeinstein: Oh, this returns `void*` because `bzrTrack_t` isn't defined in the header. Nice workaround.

components/hdw-tft/hdw-tft.c
```diff
@@ -354,7 +360,7 @@ void enableTFTBacklight(void)
             .duty       = 255, // Disable to start.
         };
         ESP_ERROR_CHECK(ledc_channel_config(&ledc_config_backlight));
-        setTFTBacklightBrightness(CONFIG_TFT_DEFAULT_BRIGHTNESS);
+        setTFTBacklightBrightness(tftBacklightIntensity);
```
AEFeinstein, line 363: If `initTFT()` is called before `setTFTBacklightBrightness()`, I think this value is undefined. That's not great. It should have a default value.

emulator/src/components/hdw-bzr/hdw-bzr.c
```diff
@@ -488,3 +488,42 @@ void bzrResume(void)
...
     }
 }
+
+/**
+ * @brief Save the state of the buzzer so that it can be restored later, perhaps
+ * after playing a different sound.
+ *
+ * @return A void-pointer which can be passed back to bzrRestore()
+ */
+void* bzrSave(void)
```
AEFeinstein, line 498: Next year I have to do a better job pulling common buzzer code out of `components`.

components/hdw-tft/hdw-tft.c
AEFeinstein: @dylwhich if I'm reading this right, it needs a little fix.

dylwhich: Ah yeah, I think so. Would it be terrible to load it from NVS here? Or I can just give the variable a default.

The actual issue I want to fix is that we call `readAllSettings()` pretty much first thing on boot, which tries to set some settings that just get overwritten by initializing the components right after. I tried to just call `readAllSettings()` again after everything was initialized, but in the case of LED brightness that still produced a noticeable flash.

AEFeinstein: Ah, if that's the issue, then the various initializers should get passed the settings from `app_main()`. It's cleaner than creating a dependency on NVM in all configurable components. And maybe `readAllSettings()` should stop setting stuff before it's initialized...

AEFeinstein: Yeah, oof, I really punted `readAllSettings()`. It should _only_ read, and pass:
* Volumes into `initBuzzer()`
* Brightness into `initTFT()`
* Brightness into `initLeds()`

Alternatively, it could read only, then have `app_main()` set settings after initializing components, but I dunno if that would solve your noticeable flash.

dylwhich: Oh wait, I actually started adding the settings as parameters. I just forgot to actually use `brightness` here, oops! I'll check on the other ones though.

dylwhich: I think that should do it for the rest of them!

AEFeinstein (approved):

PR #102, opened 2023-10-20
Switch to static i2c + add king donut example

cnlohr:
This doesn't change user-facing functionality, but rather switches the backend functionality + fixes a bug with initialization. 

It also adds a demo king donut.


AEFeinstein (changes requested):
Need to fix emu build before merging. Formatting would be nice too. Donut in the accel test mode instead of bunny maybe?

AEFeinstein (approved):

PR #105, opened 2023-10-26
Reorganize IMU and quaternion math functions

dylwhich:
### Description

Standalone math functions for quaternions, vectors, etc. are moved from `hdw-imu.c` into a separate file that can be copied wholesale between the swadge and the emulator.

### Test Instructions

Run the accel test on both the emulator and the swadge, the math should at least match up now even if the emulator doesn't really do anything.

### Readiness Checklist


AEFeinstein (approved):
This make the best of an icky situation. There isn't a way to have real and emulated components share code. Maybe that gets a makefile tweak next year to explicitly compile specific files from real components into the emulator.

PR #120, opened 2023-11-15
[Lumberjack] Fix more memory issues

AEFeinstein:
Use charactersArray in lumberjack_t
Don't use lumv after free


MrTroyD (approved):
Let's clean up this mess!

PR #127, opened 2023-11-18
Add setting option to use the swadge rotated 180 degrees

dylwhich:
### Description

Adds a "Reverse Thrust!" option in the menu which lets you use the swadge rotated 180 degrees! Not sure if this is something we want, but it turned out to be easier than I thought, so why not?

### Test Instructions

Test on hardware. Make sure that the buttons, screen, touchpad, and accelerometer all follow the rotation in a logical way.

### Readiness Checklist

### DIFF
--- a/components/hdw-btn/hdw-btn.c
+++ b/components/hdw-btn/hdw-btn.c
@@ -42,6 +42,8 @@ static int numTouchPads;
 static touch_pad_t* touchPads;
 // Used in getBaseTouchVals() to get zeroed touch sensor values
 static int32_t* baseOffsets = NULL;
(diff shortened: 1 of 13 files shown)


AEFeinstein:
> Not sure if this is something we want

I don't think anyone asked for it, so I'm gonna leave this one as an unmerged branch (but also not delete it). It's an interesting concept to revisit at some point, and a 90 degree rotation would be interesting for games that want to be in portrait mode...

dylwhich:
Fair enough!  At some point I was just thinking about the debate about holding orientation, and realized that "up" is just like, convention, man. Other than the silkscreen...

PR #138, opened 2023-11-19
Steeringwheel + Logging

cnlohr:
* Move some quaternion code into library
* Add code example of how to compute steering wheel angle.


main/modes/flight/flight.c
```diff
@@ -666,6 +666,7 @@ static void flightStartGame(flightModeScreen mode)
     flight->wintime           = 0;
     flight->speed             = 0;
 
+    accelSetRegistersAndReset();
```
AEFeinstein, line 669: In hindsight, this should probably be called by the system, not individual modes.

AEFeinstein (approved):

PR #152, opened 2023-11-22
Last few known fixes for flightsim

cnlohr:
* fix menu selection
* update to gplord's new ship
* bring over obj files from previous years
* fix vector / rotation calculation for allowing for HPR using old network protocol.

This concludes all known fixes for flightism that are required.


AEFeinstein (approved):
I'll `make format` all of `main` once code is frozen

PR #198, opened 2024-01-16
Fix flight sim for portal

cnlohr:
Fix null reference that sometimes happens when entering flight sim mode if you aren't first to the party (only happens about 10% of the time) but if near the portal will happen close to 95+% of the time.

Also, use a new slot for text as to not override what text already was, since the existing swadges squawk out an annoying message all the time.

### DIFF
--- a/main/modes/flight/flight.c
+++ b/main/modes/flight/flight.c
@@ -762,7 +762,6 @@ static void flightUpdate(void* arg __attribute__((unused)), int64_t elapsedUs)
         case FLIGHT_MENU:
         {
             drawMenuLogbook(flight->menu, flight->menuLogbookRenderer, elapsedUs);
-            drawText(&flight->logbook, c542, "rB", 237, 18);
             break;
         }
         case FLIGHT_FREEFLIGHT:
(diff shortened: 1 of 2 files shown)


main/modes/flight/flight.c
```diff
@@ -762,6 +762,7 @@ static void flightUpdate(void* arg __attribute__((unused)), int64_t elapsedUs)
         case FLIGHT_MENU:
         {
             drawMenuLogbook(flight->menu, flight->menuLogbookRenderer, elapsedUs);
+            drawText(&flight->logbook, 206, "rB", 237, 1823);
```
AEFeinstein, line 765: What is this line? It seems like it draws two chars waaaay offscreen. Also `206` shouldn't be used as a color value. The equivalent color is `c542`
```c
int16_t drawText(const font_t* font, paletteColor_t color, const char* text, int16_t xOff, int16_t yOff)
```

AEFeinstein (changes requested):

AEFeinstein (approved):

PR #201, opened 2024-01-22
Initial MacOS Support

GEMISIS:
### Description

[Problem]
Attempting to build on a MacOS device currently fails due to a variety of issues. This includes:

- Referencing "malloc.h"
- Missing checks for Apple compiler flags
- Incompatible compiler and linker flags for MacOS' clang gcc version
- The ARRAY_SIZE macro not being compatible with the MacOS' clang gcc version
- Incompatible sound library (asound)

[Solution]
This change fixes the above issues and updates the documentation for how to use on MacOS. Sound currently is a work in progress as a new sound library is searched for to use.

### Test Instructions

Built and confirmed that the simulator runs in XQuartz now, ensuring that a few different games are up and running.

### Readiness Checklist
- [X] I have run `make format` to format the changes
- [X] I have compiled the firmware and the changes have no warnings
- [X] I have compiled the emulator and the changes have no warnings
- [X] I have run `make cppcheck` and checked that `cppcheck_result.txt` has no warnings for the changes
- [X] I have added doxygen comments to any code used by more than one Swadge mode. This includes `/*! \file` comments with Design Philosophy, Usage, and Example sections for new headers.
- [X] I have run `make docs` and checked that `doxy_warnings.txt` has no warnings for the new code

### DIFF
--- a/emulator/src/rawdraw_sf.h
+++ b/emulator/src/rawdraw_sf.h
@@ -5329,7 +5329,9 @@ void AndroidSendToBack( int param )
 
 #include <stdio.h>
 #include <stdlib.h>
+#ifndef __APPLE__
 #include <malloc.h>
+#endif
 
 #ifdef HAS_XINERAMA
 	#include <X11/extensions/shape.h>
--- a/emulator/src/rawdraw_sf.h
+++ b/emulator/src/rawdraw_sf.h
@@ -565,7 +565,6 @@ static inline void DumpObjectClassProperties( jobject objToDump )
 
 
 #include <stdint.h>
-#include <malloc.h>
 
 typedef struct {
     uint32_t state[5];
@@ -3159,7 +3158,6 @@ void CNFGHandleInput()
 
 #include <windows.h>
 #include <stdlib.h>
-#include <malloc.h> //for alloca
 #include <ctype.h>
 
 HBITMAP CNFGlsBitmap;
@@ -5329,9 +5327,6 @@ void AndroidSendToBack( int param )
 
 #include <stdio.h>
 #include <stdlib.h>
-#ifndef __APPLE__
-#include <malloc.h>
-#endif
 
 #ifdef HAS_XINERAMA
 	#include <X11/extensions/shape.h>
@@ -6308,11 +6303,6 @@ OTHER DEALINGS IN THE SOFTWARE.
 
 #ifdef _CNFG_FANCYFONT
 #include "TextTool/FontData.h"
-#endif
-
-//TODO: Refactor to remove malloc reliance.
-#ifndef __clang__
-#include <malloc.h>
 #endif
 
 int CNFGPenX, CNFGPenY;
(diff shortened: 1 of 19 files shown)


main/modes/flight/flight.c
AEFeinstein, line 1: Thank you for running `make format`. We're officially friends now!

AEFeinstein:
I'm wondering if we need `<malloc.h>` at all or if `<stdlib.h>` would suffice...

GEMISIS:
> I'm wondering if we need `<malloc.h>` at all or if `<stdlib.h>` would suffice...

My understanding is we may not need it at all anymore, but I'm not near a Windows and Linux machine to confirm.

emulator/src/rawdraw_sf.h
```diff
@@ -5329,7 +5329,9 @@ void AndroidSendToBack( int param )
 
 #include <stdio.h>
 #include <stdlib.h>
+#ifndef __APPLE__
 #include <malloc.h>
+#endif
```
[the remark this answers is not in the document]
GEMISIS: Oh cool! Yeah should be a fairly straightforward change, I can take a peek at upstreaming this in that case.

main/utils/macros.h
```diff
@@ -51,7 +45,26 @@
  * @param arr An array to find the number of elements in
  * @return The the number of elements in an array (not the byte size!)
  */
-#define ARRAY_SIZE(arr) (sizeof(arr) / sizeof((arr)[0]) + STATIC_EXP(IS_ARRAY(arr)))
+#ifdef __APPLE__
+    /* Force a compilation error if condition is true, but also produce a
+       result (of value 0 and type size_t), so the expression can be used
+       e.g. in a structure initializer (or where-ever else comma expressions
+       aren't permitted). */
+    #define BUILD_BUG_ON_ZERO(e) (sizeof(struct { int : -!!(e); }))
```
GEMISIS, line 53: Can someone make sure this is okay by the way? It's [coming from Linux directly](https://github.com/torvalds/linux/blob/6f0d349d922ba44e4348a17a78ea51b7135965b1/tools/include/linux/bug.h), so I'd imagine it's fine, but it is licensed under `GPL-2.0` according to the Linux source code, so I want to make sure there's no issues here.

emulator/src/rawdraw_sf.h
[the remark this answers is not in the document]
AEFeinstein: I volunteer you!

main/utils/macros.h
AEFeinstein: I'm not a legal eagle, but everything we do is publically available, so it seems fine to use GPL2 code, especially a snippet like this.

AEFeinstein:
> I'm wondering if we need <malloc.h> at all or if <stdlib.h> would suffice...
>
> My understanding is we may not need it at all anymore, but I'm not near a Windows and Linux machine to confirm.

If you remove it, I can try building your fork on win & lin. Just go for it!

AEFeinstein (approved):
This all looks good to me. I'm gonna wait to have @JVeg199X try building this on his Mac to make sure its nice and replicable, then pull it in (and maybe clean up some of those `malloc.h` myself later...)

AEFeinstein (approved):
Windows and Linux both build emulator and firmware just fine without `malloc.h` so we're good to go here! I'm OK with `rawdraw_sf.h` diverging for now, but those changes should be unstreamed at your earliest convenience. We can always poke @cnlohr to approve it faster 😃

PR #225, opened 2024-05-12
Macos dev fixes

JVeg199X:
### Description

Made all the necessary fixes and changes to be able to launch and debug the Swadge Emulator on macOS from the comfort of your own VS Code.

Previously you needed to manually launch the Swadge Emulator from a separate XQuartz terminal window. Also, it was not possible to debug. 

Follow the revised instructions in the documentation and this should be handled automatically with the `(lldb) Launch` configuration, except for the step to launch the pulseaudio service.\.

### Test Instructions

clean emulator
Run without Debuggin (with `(lldb) Launch` configuration selected)

### DIFF
--- a/emulator/src/extensions/keymap/ext_keymap.c
+++ b/emulator/src/extensions/keymap/ext_keymap.c
@@ -10,7 +10,7 @@
 //==============================================================================
 
 static bool keymapInit(emuArgs_t* emuArgs);
-static int32_t keymapKeyCb(uint32_t keycode, bool down);
+static int32_t keymapKeyCb(uint32_t keycode, bool down, modKey_t modifiers);
 
 //==============================================================================
 // Structs
@@ -52,7 +52,7 @@ const emuExtension_t keymapEmuCallback = {
     .fnInitCb        = keymapInit,
     .fnPreFrameCb    = NULL,
     .fnPostFrameCb   = NULL,
-    .fnKeyCb         = keymapKeyCb,
+    .fnKeyCb         = &keymapKeyCb,
     .fnMouseMoveCb   = NULL,
     .fnMouseButtonCb = NULL,
     .fnRenderCb      = NULL,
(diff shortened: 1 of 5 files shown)


emulator/src/extensions/keymap/ext_keymap.c
```diff
@@ -100,7 +100,7 @@ static bool keymapInit(emuArgs_t* emuArgs)
     return false;
 }
 
-static int32_t keymapKeyCb(uint32_t keycode, bool down)
+static int32_t keymapKeyCb(uint32_t keycode, bool down, modKey_t modifiers)
```
AEFeinstein, line 103: oops. we should pay more attention to warnings from the emulator...
```c
/**
 * @brief Function to be called whenever a key is pressed.
 *
 * To allow the key press to be sent to other callbacks and the emulator, return 0.
 * To consume the key press and stop it from being sent to the emulator, return a negative value.
 * To replace the key press with a new one, return the new key code.
 *
 * @param keycode The key code. This is either a lowercase ASCII value or a \c CNFG_KEY_x code defined in CNFG.h
 * @param down true if the key was pressed, or false if the key was released
 * @param modifiers A bitfield representing all the modifier keys currently held down
 * @return A new keycode to replace the event with, or -1 to cancel it, or 0 to do nothing.
 */
int32_t (*fnKeyCb)(uint32_t keycode, bool down, modKey_t modifiers);
```

emulator/src/extensions/keymap/ext_keymap.c
```diff
@@ -52,7 +52,7 @@ const emuExtension_t keymapEmuCallback = {
     .fnInitCb        = keymapInit,
     .fnPreFrameCb    = NULL,
     .fnPostFrameCb   = NULL,
-    .fnKeyCb         = keymapKeyCb,
+    .fnKeyCb         = &keymapKeyCb,
```
AEFeinstein, line 55: I don't think you need the `&` here. Or if you do, wouldn't you need it for `keymapInit` as well?

AEFeinstein (commented):
This looks good, aside from the one suspect `&`.

I'm going to wait until https://github.com/cntools/rawdraw/pull/109 and https://github.com/cntools/cnfa/pull/24 are accepted and the submodule hashes are updated in this branch before merging to `main`.

It would be nice if https://github.com/cntools/cnfa/pull/23 gets merged as well, but not necessary.

emulator/src/extensions/keymap/ext_keymap.c
AEFeinstein: Removed it.

AEFeinstein (approved):

Issue #219, opened 2024-05-05
Themed Menu


AEFeinstein:
Make a themed menu that looks like this:

![image](https://github.com/AEFeinstein/Super-2024-Swadge-FW/assets/231180/42aaf7a6-967a-45da-a543-fca653210c8f)

The fonts are Righteous with a 150% vertical stretch and Rodin Pro Extra Black

### TODO

- [ ] Animate the background
- [ ] Offset text from parallelograms & outline it
- [ ] Fix battery indicator placement
- [ ] Clean up fonts
    - [ ] Fix vertical font offsets
- [ ] LED Effects

AEFeinstein:
Progress in https://github.com/AEFeinstein/Super-2024-Swadge-FW/tree/theme-menu

TODO

- [ ] Animate the background
- [ ] Offset text from parallelograms & outline it
- [ ] Fix battery indicator placement
- [ ] Clean up fonts
    - [ ] Fix vertical font offsets
- [ ] LED Effects

PR #226, opened 2024-05-18
Font outline

AEFeinstein:
### Description

Added a way to create outline fonts, which can be drawn over an existing font to outline it.

Add a way to have the mania renderer allocate and free its own fonts.

### Test Instructions

The outline font is used in the mania menu renderer.

### Ticket Links

#219

### Readiness Checklist


dylwhich:
Looks good to me, except the item font here seems a little weird (the lowercase `a` and `e` get chopped a bit) - is this just because the outline is the same color as the background?
screenshot-1716166362242

dylwhich:
> is this just because the outline is the same color as the background?

Wait, I can just try it and see! 
screenshot-1716166971599

Yeah, it looks fine with a non-background color even though the thin letters are still technically chopped up. So I think it's a non-issue for this PR.

AEFeinstein:
Yeah, I made the decision to draw the outline on this inside edge of the font rather than the outside edge so that the outline font would be the exact same size and could be drawn without any offsets. The downside is that thin fonts don't take it so great.

AEFeinstein:
Closed in favor of #236

PR #272, opened 2024-09-07
Create snapshot release with downloads when merging into main

dylwhich:
### Description

* The CI process was updated to automatically create a new github release called "snapshot" whenever changes are merged to main.
* The Windows build now uses static `pthread` instead of linking against the shared `libwinpthread-1`, so including `libwinpthread-1.dll` in the artifact is no longer necessary!
* Minor updates to documentation to reflect changes.
* Finally using msys2 action with caching for Windows (I ended up switching the order of the firmware / emulator as suggested). Now it's almost a whole minute faster!

### Test Instructions

Here's an example from my fork of what this will create: https://github.com/dylwhich/Swadge-IDF-5.0/releases/tag/snapshot.

Before merging, you can download the artifacts from the action (or from the last run on my fork) and check that all the files are there. You should be able to just unzip and double-click on any platform now!

Check the firmware to make sure everything is there.

### Readiness Checklist

### DIFF
--- a/.github/workflows/build-firmware-and-emulator.yml
+++ b/.github/workflows/build-firmware-and-emulator.yml
@@ -24,10 +24,10 @@ jobs:
             family: windows
             emulator: true
             firmware: true
-            emu_artifacts: |
-              swadge_emulator.exe
-              version.txt
-              C:\msys64\mingw64\bin\libwinpthread-1.dll
+            emu_artifacts:
+              - swadge_emulator.exe
+              - version.txt
+              - C:\msys64\mingw64\bin\libwinpthread-1.dll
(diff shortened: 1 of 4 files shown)


AEFeinstein:
I feel like there's a joke about double-zipping executables somewhere... It's silly, but a price I'm willing to pay.

There are a few CI warnings (mostly OSX), but I'm guessing you saw them and they don't matter?
image

AEFeinstein (approved):

dylwhich:
Ah yeah I think those errors are just like, we're installing stuff unnecessarily, which I think should be fine in case something isn't installed by default in the future.

PR #283, opened 2024-09-18
Emulator MIDI File Opening

dylwhich:
### Description

Adds the ability to open a MIDI file with `--midi-file <filename>`, which starts the emulator in the `MIDI Player` mode and does some magic to inject the MIDI File contents into a fake CNFS entry. You can also omit the `--midi-file` argument and pass the MIDI file as the last argument, which should make dragging a MIDI file onto the executable Just Work to open things on Windows at least.

### Test Instructions

- Use `--midi-file` to start the emulator with a specific MIDI file
- Try to open a nonexistent file with `--midi-file` and check that the error message makes sense
- Try dragging a MIDI file onto the executable (on Windows)

### Readiness Checklist

### DIFF
--- a/emulator/src/components/hdw-nvs/hdw-nvs.c
+++ b/emulator/src/components/hdw-nvs/hdw-nvs.c
@@ -17,6 +17,7 @@
 #include "cJSON.h"
 #include "emu_main.h"
 #include "emu_utils.h"
+#include "hashMap.h"
 
 //==============================================================================
 // Defines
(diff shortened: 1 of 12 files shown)


emulator/src/components/hdw-nvs/hdw-nvs.c
```diff
@@ -29,6 +30,22 @@
...
+
+typedef struct
+{
+    char key[36];
+    uint32_t size;
+    bool isBlob;
+    union
+    {
+        void* blob;
+        int32_t int32;
```
AEFeinstein, line 45: I haven't read what this does yet, but on most non-embedded systems an `int32_t` is half the size of a `void*`.

emulator/src/emu_cnfs.c
AEFeinstein, line 1: Switching from SPIFFS to CNFS was nice to not have to mock the functions in the emulator. Well, it was nice while it lasted. I don't see a better way.

emulator/src/components/hdw-nvs/hdw-nvs.c
AEFeinstein: Ah, I see. All good.

AEFeinstein (approved):

Issue #299, opened 2024-09-27
Improve MIDI Playback Performance


dylwhich:
**Summary**
The MIDI engine hasn't been optimized yet and we're running low on cycles so we should take a look

**Technical Spec**
List of random things to look into later, vaguely in order of how much I think they'll help with no real basis:

 * Move ADSR logic into sw synth, because it kind of already does this a bit anyway and will probably simplify the logic
 * Check for events less frequently, as the MIDI tick rate is vastly slower than the sample rate
 * Clean up the logic that decides whether to check for the next event. It's pretty complicated and it might be faster to just always do the check once
 * Look for 64-bit cars and math
 * Less interpolation with slightly more state
 * Check if all those `__builtin_ctz` and `__builtin_popcount` calls are actually turning into single instructions
 * Only sum oscillators for voices that are actually active (but this might allow "heavier" parts of songs to slow stuff down vs. always doing the same amount of work regardless)

AEFeinstein:
I'd also recommend using https://github.com/KDAB/hotspot to profile the emulator and see where the cycles are going. It's not the same as the firmware, but it should be close enough to find any big offenders. I used it when optimizing texture lookups for the raycaster last year.

The makefile target for generating code coverage data can also be kinda helpful for profiling a specific function be looking at the number of times each line is run

AEFeinstein:
Some idea of spots to look at
![image](https://github.com/user-attachments/assets/518f52fe-265e-4e31-a47a-a761ce446a0d)

AEFeinstein:
Good as we're getting!

Issue #278, opened 2024-09-13
Gamepad Mode


AEFeinstein:
**Summary**

Spruce up the UI. Is there anything else we'd want to change?

**Technical Spec**
_How will this be implemented?_
_What peripherals will be used, and how?_

**UI**
_What are the display screens & menus (if applicable)?_
_What user inputs are there for each screen?_

**Mockups**
_Paste any mockups of screens or anything else here._

**How to Test**
How can we verify the feature is working?

AEFeinstein:
@JVeg199X were there configs that you wanted to add here?

AEFeinstein:
Configs
* Swap touchpad between stick and virtual buttons

Also something with LEDs

AEFeinstein:
I believe this is done. @JVeg199X if it ain't, feel free to reopen.

Issue #232, opened 2024-05-22
Pango


AEFeinstein:
To be filled out by @JVeg199X 

----

**Summary**
_What is the feature all about? Is it a game, a utility, an enhancement, or something else?_

**Technical Spec**
_How will this be implemented?_
_What peripherals will be used, and how?_

**UI**
_What are the display screens & menus (if applicable)?_
_What user inputs are there for each screen?_

**Mockups**
_Paste any mockups of screens or anything else here._

**How to Test**
How can we verify the feature is working?

emilyanthony4244:
I noticed in the Pango tutorial the player is instructed to hit Start. This should be changed to match the button labels on the Swadge.

PR #313, opened 2024-11-12
A little bit of everything: Pango, Gamepad, Midi optimizaitons

JVeg199X:
### Description

Pango (#232):
- fix bug with enemy behavior that throws off block counting
- add hotdog particle and bonus item
- start adding music (this is definitely unfinished)
- Use drawWsgTile to increase FPS

Midi (#299):
- Don't sum oscillators for inactive voices or when current volume and target volume are both 0
- Always check for midi events at least once

Graphics code:
- Set explicit RGB value for cTransparent (black)

Gamepad (#278):
- Start adding an option for Switch gamepad to use touchpad as extra buttons


emulator/src/components/hdw-tft/hdw-tft.c
```diff
@@ -280,7 +280,7 @@ void drawDisplayTft(fnBackgroundDrawCallback_t fnBackgroundDrawCallback)
                     int colorolor = frameBuffer[(y * TFT_WIDTH) + x];
                     if (colorolor == 216)
                     {
-                        colorolor = c500;
```
AEFeinstein, line 283: This was intentional to highlight when transparent pixels are drawn, because they shouldn't be. Maybe I'm not understanding the use case, but shouldn't wsgs with transparency be drawn with transparent functions, and vice versa?

main/midi/midiPlayer.c
```diff
@@ -1198,23 +1255,15 @@ int32_t midiPlayerStep(midiPlayer_t* player)
...
-                <= SAMPLES_TO_MIDI_TICKS(player->sampleCount, player->tempo, player->reader.division))
-            {
-                checkEvents = true;
-            }
-        }
-        else
+        if (!player->eventAvailable)
         {
             ESP_LOGI("MIDI", "Done playing file!");
             midiSongEnd(player);
```
dylwhich, line 1269: `checkEvents` should be set to `false` here once the song has ended so it doesn't keep trying to check events later.

emulator/src/components/hdw-tft/hdw-tft.c
AEFeinstein: And this is specific to the emulator, so this is supposed to be a visual warning. Does the fact that you added an explicit value for `cTransparent` to `paletteColors[]` make this wholly unnecessary?

AEFeinstein:
@JVeg199X you're lucky I'm not one of those draconian PMs who enforce one-ticket-per-branch! There's a little feedback to go through here and then we're good to merge.

emulator/src/components/hdw-tft/hdw-tft.c
JVeg199X: The idea is to gain some FPS by using `drawWsgTile` to draw the tilemap instead of `drawWsgSimple`. 

All the tiles were designed with transparent pixels, which when normally drawn with `drawWsgTile`, appear as an off-white color. But, since the background in Pango is always black, we can make it look correct by adding an explicit definition for `cTransparent`, setting it to black.

The emulator handles `cTransparent` separately in this `if` statement here, but I should probably just add an explicit definition to the emulator palette instead...

main/modes/games/pango/pango.c
```diff
@@ -259,6 +259,9 @@ void pangoEnterMode(void)
     pango->tilemap.entityManager    = &(pango->entityManager);
     pango->tilemap.tileSpawnEnabled = true;
 
+    pa_loadMapFromFile(&(pango->tilemap), "preset.bin");
+    pa_generateMaze(&(pango->tilemap));
```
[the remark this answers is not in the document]
JVeg199X: The same algorithm used in the actual Pengo game, minus bugs!

https://www.ukvac.com/forum/threads/sega-pengo-bug-found-in-maze-generating-algorithm.23920/

main/midi/midiPlayer.c
JVeg199X: Fixed

main/modes/games/pango/pango.c
```diff
@@ -394,6 +397,7 @@ void pangoMainLoop(int64_t elapsedUs)
     }
 
     pango->update(pango, elapsedUs);
+    DRAW_FPS_COUNTER(pango->gameData.scoreFont);
```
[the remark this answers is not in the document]
JVeg199X: Yup!

emulator/src/components/hdw-tft/hdw-tft.c
JVeg199X: Of course, if we can further optimize the MIDI playback, I can go without the `cTransparent` changes.

AEFeinstein: As long as it's consistent and expected, the change is fine. I'd add to the emulator palette and call it a day.

main/modes/games/pango/pango.c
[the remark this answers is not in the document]
AEFeinstein: Wow, love that accuracy! Last time I generated mazes I used Eller's.
https://weblog.jamisbuck.org/2010/12/29/maze-generation-eller-s-algorithm

emulator/src/components/hdw-tft/hdw-tft.c
AEFeinstein: I'll add it to the emulator palette rn actually

AEFeinstein (approved):

PR #339, opened 2024-12-15
Bigbug

DebrisHauler:
### Description

There are various different songs happening at different game states. Hurry up music, fail & clone at home music, dive summary music. LED light animations got overhauled for the latest LED layout. Bugs are walking and flying around each with their own speed. You can press the pause button to look at your radar (map). You can tether up to nearby bugs if they are a physics object (dead; shot 5 times). Bugs can be dragged to the booster's grabby hand. Or it will just grab unsuspecting bugs that travel there. Every 10 bugs creates a mega booster ping which will give you a random choice from a pool of radar upgrades. When you've taken every radar upgrade, the remaining choice on repeat tops off your ammo count. There are cars and doors. Interacting with cars that don't have a door is not recommended, as that is a different objective I haven't worked out yet. When you touch a car, the alarm is triggered. A horde of bugs is spawned, the size of which depends on the depth of the car. They get pretty crazy at the bottom of the map. The door closes the arena in. The code to close the door is stupidly complex because the door may be cached when it is off screen. You can always dig your way in or out, but it's best to keep the bugs enclosed so you can get them. Once you've killed a number of bugs equal to the number spawned, then the car's trunk opens and the door opens. The car trunk will give you something to improve Garbotnik, but this is not implemented yet.

### Test Instructions

How was this tested? Painfully. There are some crashes going on if you lift off or die, and maaaaaybe sometimes when you start a car alarm. Not sure if that one is fixed or not.

### Readiness Checklist
- [👎] I have run `make cppcheck` and checked that `cppcheck_result.txt` has no warnings for the changes


AEFeinstein:
It's warning about `bb_EnterMode()`, but that's expected.

cppcheck is throwing warnings about my background drawing code, but I think those are false positives.

AEFeinstein (approved):

PR #380, opened 2025-01-07
Joystick input for Windows & Linux

dylwhich:
### Description

Joystick inputs! They work on Windows and Linux, I haven't attempted OSX at all.

On both platforms, the joystick must be connected before the emulator starts. If the joystick is disconnected, the emulator will not attempt to reconnect to it.

Also in support of arcade-y things, I fixed the emulator display not being centered properly in fullscreen and some other situations that come up more often now that the LEDs have moved.

### Test Instructions

It's more or less plug and play! On Windows, will open the first joystick device available. On Linux, will open `/dev/input/js0`.

1. Plug in a swadge to USB
2. Open up Gamepad mode and start in Computer mode
3. Start the emulator
4. It should Just Work!
5. Try the Touch Test and Accel Test modes, and also everything else.
6. Open the emulator in fullscreen

### Readiness Checklist

### DIFF
--- a/emulator/src/extensions/gamepad/ext_gamepad.c
+++ b/emulator/src/extensions/gamepad/ext_gamepad.c
@@ -8,11 +8,9 @@
 
 #include <esp_log.h>
 
-#include <fcntl.h>
 #include <stdio.h>
 #include <inttypes.h>
 #include <stdlib.h>
-#include <unistd.h>
 #include <errno.h>
 #include <math.h>
 #include <stdint.h>
@@ -22,9 +20,9 @@
 #include <joystickapi.h>
 #elif defined(EMU_LINUX)
 #include <linux/joystick.h>
+#include <fcntl.h>
+#include <unistd.h>
 #elif defined(EMU_MACOS)
-#else
-#error "Unrecognized platform"
 #endif
 
 // Defines the mapping of the default Swadge Gamepad mode
@@ -84,8 +82,6 @@ typedef struct
     bool pendingState;
     // Whether the second POV hat axis is pending
     bool pendingPov;
-    int* axisMins;
-    int* axisMaxs;
 } emuWinJoyData_t;
 #endif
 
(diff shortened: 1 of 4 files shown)


emulator/src/extensions/gamepad/ext_gamepad.c
```diff
@@ -0,0 +1,628 @@
...
+                // If we're here, we have successfully transmitted all state changes via events
+                // Now, try to get the new joystick state and see if it's different
+
+                // 1. Get the current joystick state data
+                // 2. Check which values are different from the last state
+                // 3. Emit the event with the new data
+                // 4. Update the last state to match the new data returned
+                // 5. If nothing is different, set pendingState = false;
+                JOYINFOEX winEvent = {0};
+                // what the hell kind of API requires this
```
AEFeinstein, line 270: Windows API!

emulator/src/extensions/gamepad/ext_gamepad.c
```diff
@@ -0,0 +1,628 @@
...
+                else
+                {
+                    // nothing is different!
+                    // check one more time
+                    winData->pendingState = false;
+                    continue;
+                }
+            }
+        } while (0);
+#elif defined(EMU_LINUX)
```
AEFeinstein, line 431: Seems nicer

emulator/src/extensions/gamepad/ext_gamepad.c
```diff
@@ -0,0 +1,628 @@
...
+
+                        emulatorSetAccelerometer(accelX, accelY, accelZ);
+                        break;
+                    }
+
+                    case DPAD_AXIS_X:
+                    case DPAD_AXIS_Y:
+                    {
+                        buttonBit_t curState = 0;
+                        if (joystickExt.axisData[DPAD_AXIS_X] < 0)
```
AEFeinstein, line 578: So on the actual Swadge the D-Pad is four discrete buttons, but here's it's a proper D-Pad, so the hardware can handle simultaneous up/down presses while the joystick cannot. I don't think it matters, just a point of interest.

AEFeinstein (approved):

emulator/src/extensions/gamepad/ext_gamepad.c
dylwhich: Ah, yeah that's true, probably not something we'd realistically need to support anyway. This is also just how the swadge maps the D-pad in gamepad mode, and I guess it's pretty standard since a switch controller also maps its D-pad the same way.

emulator/src/extensions/gamepad/ext_gamepad.c
dylwhich: I think so too but in fairness to windows I wrapped the linux API pretty tightly and just made the windows one comply with that which looks a bit more ugly lol

PR #381, opened 2025-01-07
More Console Commands and Replay Script Support

dylwhich:
### Description

* Adds the option to only fuzz certain buttons when fuzzing
* Adds a bunch of console commands to the emulator, mainly ones to control fuzzing
* Adds the ability to include a console command within a replay script CSV
* Documents all the existing and new console commands
* Adds a slightly more advanced help to the console
* Fixes a bug with the emulator where hiding and showing panes on the fly didn't work properly

### Test Instructions

Try some console commands, maybe try including a console command in a replay script.

### Readiness Checklist

### DIFF
--- a/emulator/src/extensions/fuzzer/ext_fuzzer.c
+++ b/emulator/src/extensions/fuzzer/ext_fuzzer.c
@@ -29,6 +29,8 @@ typedef struct
     bool touch;
     bool motion;
     bool time;
+
+    buttonBit_t buttonMask;
 } fuzzer_t;
 
 //==============================================================================
@@ -60,6 +62,8 @@ static bool fuzzerInitCb(emuArgs_t* emuArgs)
     fuzzer.motion  = emuArgs->fuzzMotion;
     fuzzer.time    = emuArgs->fuzzTime;
 
+    fuzzer.buttonMask = (PB_A | PB_B | PB_UP | PB_DOWN | PB_LEFT | PB_RIGHT | PB_START);
+
     if (emuArgs->fuzz)
     {
         printf("\nFuzzing:\n - [%c] Buttons\n - [%c] Touch\n - [%c] Motion\n - [%c] Time\n", fuzzer.buttons ? 'X' : ' ',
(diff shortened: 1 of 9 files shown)


emulator/src/extensions/fuzzer/ext_fuzzer.c
```diff
@@ -60,6 +62,8 @@ static bool fuzzerInitCb(emuArgs_t* emuArgs)
     fuzzer.motion  = emuArgs->fuzzMotion;
     fuzzer.time    = emuArgs->fuzzTime;
 
+    fuzzer.buttonMask = (PB_A | PB_B | PB_UP | PB_DOWN | PB_LEFT | PB_RIGHT | PB_START);
```
AEFeinstein, line 65: I think in the past we used `--lock` when fuzzing to avoid the `PB_SELECT` problem. This works too, and opens the opportunity to make it configurable maybe one day.

emulator/src/extensions/replay/ext_replay.c
```diff
@@ -57,9 +57,10 @@ typedef enum
     SCREENSHOT,
     SET_MODE,
     RANDOM_SEED,
+    COMMAND,
 } replayLogType_t;
 
-#define LAST_TYPE RANDOM_SEED
+#define LAST_TYPE COMMAND
```
AEFeinstein, line 63: You could add `NUM_REPLAY_LOG_TYPES` as the final enum value and then use that as your loop bound. It saves you having to update `LAST_TYPE` if/when you add more commands.

AEFeinstein (approved):

emulator/src/extensions/fuzzer/ext_fuzzer.c
dylwhich: Yeah, the feedback I got from Thaeli that prompted this was that these days `--lock` doesn't really help fuzz a specific mode much, the fuzzer ends up spending most of its time opening and closing quick settings and changing the brightness instead.

PR #398, opened 2025-01-16
Fix TTT CPU Difficulty Selector

dylwhich:
### Description

Fixes the difficulty being lost when the CPU player data is reset.

### Test Instructions

Try hard mode and see if it's actually hard.

### Readiness Checklist


dylwhich:
I got suspicious the debugging stuff I tried to draw never showed up. And the printfs didn't work. And yet it still took me a couple hours to figure out that `make clean` was never going to fix it...

Well, I already wrote the debugging stuff so I think I'll try to slightly improve the AI still, it's got a couple issues I already noticed. So I don't forget: If there are two subgames that would both win the game if won, the AI will pick the first one, even if the other one can be won immediately (and win the entire game).

AEFeinstein (approved):

PR #402, opened 2025-01-17
Configurable joystick

dylwhich:
### Description

Adds a bunch of commands to configure the joystick mappings and a system for presets. Also fixes a broken markdown table (who knew escaping `|` was so complicated!) in the docs.

### Test Instructions

If you have a switch pro controller, you can test out the preset, just start the emulator with it connected and run the emulator with `--preset switch`.

For the commands, open the emulator, hit F4, and type `help joystick` which should list a bunch of commands you can use. The descriptions should be clear enough, and they're also documented in `EMULATOR.md`. If you have a non-swadge joystick, you should be able to set it up using https://hardwaretester.com/gamepad - all command arguments that take an axis or a button are 0-indexed.

### Readiness Checklist


dylwhich:
actually let me add a 360 controller mapping...

dylwhich:
nevermind, don't think I'm going to figure that one out without actually having one

emulator/src/extensions/tools/emu_console_cmds.c
```diff
@@ -40,12 +38,22 @@ static const char* commandDocs[][3] = {
      "begin recording inputs into replay file [filename], or an auto-generated file name if not specified"},
     {"fuzz", "fuzz [on|off]", "toggles the fuzzer"},
     {"fuzz buttons", "fuzz buttons [on|off]", "toggles fuzzing of button inputs"},
-    {"fuzz buttons mask", "fuzz buttons mask [<button>[ <button>  ...]]",
+    {"fuzz buttons mask", "fuzz buttons mask [<button> [<button>  ...]]",
```
AEFeinstein, line 41: When doing bigbug fuzzing, I tweaked it so that the buttons had different randomized weights, and the weights changed over time. Basically this would move Garbotnik in different directions rather than have him twitch in place forever. That doesn't feel feasible as CLI, but might be doable as JSON config? I'm not sure how valuable it would be, as the magic of fuzzing is that you usually do it without thinking too much...

AEFeinstein (approved):

emulator/src/emu_utils.c
```diff
@@ -218,3 +218,46 @@ const char* getTimestampFilename(char* dst, size_t n, const char* prefix, const
...
+    }
+
+    return (buttonBit_t)0;
+}
+
+const char* getButtonName(buttonBit_t btn)
+{
+    for (int i = 0; i < 8; i++)
+    {
+        if (btn & (1 << i))
```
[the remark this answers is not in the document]
dylwhich: The idea here was just to output _something_ in the event that `btn` has multiple bit set, though nothing should be calling this with multiple bits set at the moment.

docs/EMULATOR.md
```diff
@@ -89,6 +126,8 @@ Emulates a swadge
      --fuzz-motion[=y|n]     Set whether motion inputs are fuzzed
      --headless              Runs the emulator without a window.
      --hide-leds             Don't draw simulated LEDs next to the display
+ -j, --joystick=JOYDEV       Sets the joystick device to use.
```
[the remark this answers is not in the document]
dylwhich: Oh cool! This reminds me that I want to look into adding a USB profile for controlling the LEDs, I think there's like a HID LED profile or something. I'll make a ticket!

emulator/src/extensions/tools/emu_console_cmds.c
dylwhich: Ah, this is a pretty good idea. I think there actually was a little bit of configuration for setting probability on the 2023(?) emulator but yeah something a bit better than completely random button presses would be good. The other thing is it's _constantly_ pressing buttons but never for very long which yeah usually means not much happens. It wouldn't be too hard to make even a slightly smarter fuzzer. Even if it just spaced out button presses a bit and held them for more reasonable times that would be a big improvement.

And yeah some sort of emulator config might be a good idea, some of the parameters are getting a bit unwieldy and we've accumulated a good bit of stuff we'd actually want to configure.

emulator/src/extensions/gamepad/ext_gamepad.c
```diff
@@ -25,19 +26,90 @@
...
-    PB_A, PB_B, 0, 0, 0, 0, 0, 0, 0, 0, PB_SELECT, PB_START, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+const emuJoystickConf_t joyPresetSwitch = {
+    .name = "Switch",
+    .touchpad = {
+        .xAxis = 2,
+        .yAxis = 3,
+        .deadzone = 1024,
+    },
+
+    .accel = {
```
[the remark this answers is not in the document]
dylwhich: Hmm, I don't see anything on the `js0` interface so I think that might just be a different HID device or something. I do have some weird behavior with the switch controller where it doesn't show up as a joystick unless I plug it into a USB 2.0 port. USB-C and USB 3.0 ports just don't work, so I assume it's using something fancier in that case? Or I'm missing a kernel driver? Who knows, computers are mysterious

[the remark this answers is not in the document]
AEFeinstein: https://www.youtube.com/watch?v=fq3abPnEEGE

docs/EMULATOR.md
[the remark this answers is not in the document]
AEFeinstein: Not my intent, but go for it!

emulator/src/extensions/tools/emu_console_cmds.c
AEFeinstein: Yeah, turns out when you make features complex, usage goes down. It's a tough tradeoff.

The holy grail would be a fuzzer that had enough sense to "do something" rather than just hammer on buttons. I vaguely remember using an Android one at one point which would take screenshots and 'reward' the fuzzer for getting to different states or something like that.

AEFeinstein: Oh hey https://llvm.org/docs/LibFuzzer.html

dylwhich: > Oh hey https://llvm.org/docs/LibFuzzer.html
> Maybe we should not "roll our own" here

Oooh, wow! That's incredible, if I'm understanding that interface right then it should be trivial to hook that into the existing button input system. I might try that later.

AEFeinstein: Yep, it looks trivial to hook up to buttons, but it does recommend having deterministic targets that don't modify global state, and we have neither. Even with seeded rng, imprecise us timers make it non deterministic

dylwhich: Hmm... We do have the "fake timer" stuff that should help with that... it should give modes exactly `frameTimeUs` every frame, regardless of how long it actually took. More deterministic fuzzing is basically what it's for!

PR #403, opened 2025-01-17
Minor makefile fixes

dylwhich:
### Description

Fixes a couple things that bothered me just now.

1. Prioritize `clang-format-17` over the default one, because my new laptop is on a new fedora which comes with `clang-format` 19, which of course disagrees with 17 so I have to keep changing the makefile.
2. Add a separate makefile target for `wget plantuml.jar` instead of running it in `make docs`. Not sure why, but every time I ran `make docs` it would print a message saying plantuml.jar was already present and won't be downloaded, but then it downloaded it anyway.

### Test Instructions

Tested by running `make format` and `make docs` on several different machines with different versions of stuff. Seemed to work fine to me!

### Readiness Checklist

### DIFF
--- a/makefile
+++ b/makefile
@@ -397,8 +397,10 @@ fullclean: clean
 # Utility targets
 ################################################################################
 
-docs:
+plantuml.jar:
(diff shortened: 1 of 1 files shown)


AEFeinstein (approved):
I never noticed the plantuml thing, aside from it needing to be downloaded again after a fullclean

PR #409, opened 2025-01-20
Fix infinite loop when exiting BigBug

dylwhich:
### Test Instructions

Be in the middle of a level in big bug and try to exit. Sometimes it hangs forever

### Readiness Checklist

### DIFF
--- a/main/modes/games/bigbug/gameData_bigbug.c
+++ b/main/modes/games/bigbug/gameData_bigbug.c
@@ -206,7 +206,7 @@ void bb_freeGameData(bb_gameData_t* gameData)
     }
     while (gameData->pleaseCheck.first)
     {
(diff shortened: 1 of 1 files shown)


AEFeinstein:
This definitely affects the emulator, but would it affect firmware? I'm pretty sure we don't have a WDT, so I think it would?

AEFeinstein (approved):

dylwhich:
> This definitely affects the emulator, but would it affect firmware? I'm pretty sure we don't have a WDT, so I think it would?

I think it would? At least I've had similar bugs where the physical swadge just hangs forever. And I don't think this invokes UB or anything so I think it'll just run forever.

AEFeinstein:
Welp, sounds like a 2.1.2. I'll wait until like Tuesday night to do that (or maybe even Wednesday morning).

PR #414, opened 2025-01-27
sao game

cnlohr:
Ok this is an odd or as it's only in the sandbox area... Dunno what to say about it at this time.


AEFeinstein:
Oh, this is entirely in `sandbox_test`. I wrongly assumed you were working in a totally separate repo. Would it make more organizational sense to be a separete repo? We could always submodule it so the directory structure wouldn't change.

cnlohr:
I don't have particularly strong feelings about how things get organized.

AEFeinstein:
Yeah, the programming stuff in `test_programmer` makes sense here for sure. The `sao047b` part is a little weirder, especially when it gets to the web stuff. That's very much your independent game and hardware, and I'd rather have you in charge of PRs and such.

I dug around and found the hardware in https://github.com/cnlohr/cnhardware/tree/master/ch32v003/0.47SAOB. That would be good to link in a `README.md`, or maybe migrate both the firmware and hardware to it's own repo? I'm happy to submodule that here if having a wholly separate repo makes development more cumbersome.

tools/sandbox_test/test_programmer/sao047b/games.h
AEFeinstein, line 1: Code in a header without `IMPLEMENTATION` guards? Bold choice.

tools/sandbox_test/test_programmer/sao047b/LICENSE
```diff
@@ -0,0 +1,11 @@
+
+
+
+Blit16
+
+ISC License Copyright (c) 2018 Andrew Reece
```
AEFeinstein, line 6: Who's Andrew Reece?

cnlohr: Oh, I can get rid of that. It was a font I was playing with.

tools/sandbox_test/test_programmer/sao047b/web/subst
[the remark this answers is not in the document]
cnlohr: Ah! oops! will fix.

AEFeinstein:
Will be done in #467. @cnlohr I'm not sure what, if anything, you want to do with this branch.

Issue #418, opened 2025-02-01
Reference-counted Asset Loading


dylwhich:
**Summary**
_What is the feature all about? Is it a game, a utility, an enhancement, or something else?_

This is a firmware utility that would simplify the process of loading and unloading assets, and eliminate a source of memory leaks. This utility would allow the declaration of a loader for a particular asset type. When the loader is used to allocate a resource, it performs a lookup in the hash table and loads the asset if it is not present yet. If the asset is present, the existing asset is returned instead of allocating a new one. Each asset's reference count is incremented every time the asset is loaded, and decremented any time the asset is unloaded. If the reference count reaches 0, the asset is unloaded and removed from the hash map. All assets could also be unloaded at once, which could be done when a mode about to exit or returning to its main menu.

Benefits of the system:
* Simplifies the process of loading and unloading assets greatly. The loader takes care of deciding when to actually allocate or free resources, and all assets can be force-unloaded on request which makes it much harder to accidentally forget one.
* Allows for saving memory when assets are no longer needed. Useful for when you need a variety of assets, but don't necessarily need all of them allocated at the same time.

Drawbacks:
* A bit more memory & CPU overhead than manually managing assets
* Can't modify returned values, e.g. for palette swapping, since they're shared

**Technical Spec**
_How will this be implemented?_
_What peripherals will be used, and how?_

The system would be composed of a "loader" struct, which contains:
* A hash map to hold all currently-allocated asset information
* A pointer to a function for loading the asset
* A pointer to a function for unloading the asset
* The size of the asset's container struct (e.g., of a `wsg_t`)

Each member of the hash map would be a dynamically-allocated struct that contains:
* The container struct
* The asset's filename/key (e.g. `kd0.wsg`)
* The reference count
This struct and its data (meaning the container struct and the key string data) would all be allocated simultaneously, reducing allocation overhead slightly.

The API of this utility would consist of these functions:

* `void initLoader(loader_t* loader, size_t dataSize, loaderLoadFn_t loadFn, loaderUnloadFn_t unloadFn)` -- Initializes the loader and its hash map, and sets the loading function and container size members.
* `void deinitLoader(loader_t* loader)` -- Unloads all assets, then deinitializes the loader and its hash map, zeroing out all fields.
[...]

Issue #400, opened 2025-01-16
Asset Preprocessor Improvement


dylwhich:
**Summary**
Improvements to the asset preprocessor to make it easier to create new asset types, reduce boilerplate, and generally make it more usable.

**Technical Spec**

Instead of each asset processor doing the same logic to decide on a new filename, there will be an array of config structs, sort of like what's used for the "raw" file types [here](https://github.com/AEFeinstein/Super-2024-Swadge-FW/blob/main/tools/assets_preprocessor/src/assets_preprocessor.c#L26) that just get compressed. The struct would have:
* `const char* inExt`: The extension of the file to be transformed, e.g. `png`
* `const char* outExt`: The extension of the file that will be created, e.g. `.wsg`
* `enum processType_t type`: The type of processing the file gets. Right now would just be `FUNCTION`, but maybe we'd have another for `EXEC` (to use for external tools)
* `bool (*processFn)(const char* inFile, const char* outFile)`: If type is `FUNCTION`, this would be a pointer to the function called to process the file. The in and out file paths are already set up, so no filename parsing is needed inside the handler.
* (if `EXEC` is implemented) `const char* execCmd`: The command to run for processing this file. If it exits abnormally the file is considered failed. The in and out file arguments would be populated in string arguments, so `"cp %1$s %2$s"` would be a valid one. This might be a good uniform way to handle things like the sokoban python script.

Also, it should detect conflicts. If multiple input files are mapping to the same file, that should throw a fatal error.

AEFeinstein:
Thank you for making my code suck less

dylwhich:
It's all anyone can do really, just make the code suck a little bit less every day!

dylwhich:
Another type of processor we could have, which the trophy discussion just made me think of, is one that handles all the assets of a particular type at once. In the trophy example, this would mean taking a bunch of JSON files that represent a struct, and concatenating (and sorting) them into a single array that becomes a single file in CNFS. I think this could also be useful for something like, e.g. swadge hero, where you could have a second asset processor for `.chart` files that outputs an index of filenames that can be used instead of hardcoding.

AEFeinstein:
Hm, I kinda see where you're going, but one-file-in-one-file-out is a clean paradigm. I'd wait until we have a need for concatenation before building it as a generic feature.

dylwhich:
[...]

PR #436, opened 2025-04-10
Asset enum

AEFeinstein:
### Description

Replaced string filenames with enums

### Test Instructions

Compile, run, and make sure that assets are still loaded and used correctly

### Ticket Links

#418

### Readiness Checklist


main/modes/music/usbsynth/mode_synth.c
AEFeinstein, line 1: @dylwhich this file needs some work still. I'm not entirely sure what you want to do with it, but I'm leaning towards tucking it in the secrets menu unless it becomes more user-friendly.

tools/cnfs/cnfs_gen.c
```diff
@@ -6,119 +6,219 @@
...
         int len = ftell(f);
         fseek(f, 0, SEEK_SET);
 
-        struct fileEntry* fe = entries + nr_file;
+        // Save the input file into entries[]
+        struct fileEntry* fe = &entries[nr_file];
         fe->data             = malloc(len);
         fe->len              = len;
-        int r                = fread(fe->data, 1, len, f);
+        fe->padLen           = ((len + 3) / 4) * 4;
```
AEFeinstein, line 161: Maybe the ESP will be a little happier with files being 32 bit aligned?

emulator/src/emu_cnfs.c
```diff
@@ -97,50 +95,34 @@ bool deinitCnfs(void)
     return true;
 }
 
-const uint8_t* cnfsGetFile(const char* fname, size_t* flen)
+const uint8_t* cnfsGetFile(cnfsFileIdx_t fIdx, size_t* flen)
 {
-    if (cnfsInjectedFilename && !strcmp(cnfsInjectedFilename, fname))
+    if (cnfsInjectedFilename && fIdx >= CNFS_NUM_FILES)
```
AEFeinstein, line 100: @dylwhich this may need some thinking too

AEFeinstein:
@dylwhich nudge

dylwhich:
Okay, finally taking a look at this and I thought it was broken but it turns out I needed to do a fullclean, I think, because I ended up with duplicate `WHOLE_NOTE_WSG` and `HALF_NOTE_WSG` files in `assets_image` from slightly differently named files.

emulator/src/emu_cnfs.c
dylwhich: Hmm, okay, yeah I'm not sure how to do this now that I take a closer look at it... The whole reason the "injected file" thing exists was for opening MIDI files on-the-fly and letting the player load them... by name. I guess since this really is an emulator-only thing I could just switch to injecting the file data into NVS instead, which works fine, and drop the asset injection feature altogether.

dylwhich:
I think you already tried to fix this but I'm still having issues with `cnfs_image.h` not regenerating quite how we'd want. I think ideally you could add a new file, try to reference its enum constant, and run `make` without ever getting an ugly compile error or needing to do `make clean`. There is a way I can think of to do it like that, but it involves going outside of make a bit. Most of the asset preprocessing already does the timestamp check to avoid reprocessing assets that haven't changed, so I think we could do this (kinda goes along with #400):

* Always run the `assets` target (make it an order-only prereq of CNFS_FILE, I think?)
* Have the asset preprocessor write a timestamp to a file any time it updates _any_ assets
* Have that timestamp file be a regular prereq to CNFS_FILE
* So, always run asset preprocessor, and if nothing changed, no need to regenerate the enum
* If we reeeeally wanted to get fancy we could differentiate between file contents changing and the set of files changing and only also regenerate `cnfs_image.h` when the names change but that's probably overkill...

dylwhich:
> I think you already tried to fix this but I'm still having issues with `cnfs_image.h` not regenerating quite how we'd want. I think ideally you could add a new file, try to reference its enum constant, and run `make` without ever getting an ugly compile error or needing to do `make clean`. There is a way I can think of to do it like that, but it involves going outside of make a bit.

OKAY IT KINDA WORKS PERFECTLY? Check this out, I am genuinely surprised: https://github.com/AEFeinstein/Super-2024-Swadge-FW/compare/asset_enum...fancy-asset-preprocessor#diff-beda42571c095172ab63437d050612a571d0d9ddd3ad4f2aecbce907a9b7e3d0R333

The way that branch works is (imo) the ideal scenario. You can:
1. Create a new file: `touch assets/cool_stuff.txt`
2. Edit some C file to have `cnfsFileIdx_t file = COOL_STUFF_TXT;`
3. Run `make`
4. Your file gets processed
5. `cnfs_image.h` and `cnfs_image.c` get regenerated
6. Everything else also gets recompiled like you'd expect
7. The build succeeds the first time, with no errors!
8. Run `make` again
9. Nothing happens, it's all up-to-date already!!! (well, the asset processor runs, but it doesn't actually do anything other than check file timestamps)
 can specify additional targets that get built by a rule like `target target2 &:`

It all works pretty much like I said before, but I learned about [multiple targets](https://www.gnu.org/software/make/manual/html_node/Multiple-Targets.html) in make, which I think makes it actually work. The timestamp file is a _bit_ hacky but honestly it works so well and you'll never have to do a `make clean` for asset updates ever again!

So, should I pull those changes into this branch?

AEFeinstein:
Multiple targets are neat, and it's hard to argue with "OKAY IT KINDA WORKS PERFECTLY?" Might as well open the PR!

In my flu stupor I did think about this a little and came up with a simple "what if we declare the CNFS file dependent on the assets?" Basically:
```make
ASSET_FILES = $(shell $(FIND) assets -type f)
# and
$(CNFS_FILE): $(ASSET_FILES)
    $(MAKE) -C ./tools/assets_preprocessor/
    ./tools/assets_preprocessor/assets_preprocessor -i ./assets/ -o ./assets_image/
    $(MAKE) -C ./tools/cnfs/
    ./tools/cnfs/cnfs_gen assets_image/ $(CNFS_FILE) $(CNFS_FILE_H)
```
If I understand that right, any time an asset changes, it'll regenerate the CNFS file? It seems OK, so I made that change in 2a320ffe53023b4d093c6d21bc041776fc6191fc. Before I merge your PR, could you check if that change also kinda works? I get the feeling that no matter where we land, its gonna break on someone's machine and we'll patch it later anyway.

dylwhich:
> If I understand that right, any time an asset changes, it'll regenerate the CNFS file? It seems OK, so I made that change in [2a320ff](https://github.com/AEFeinstein/Super-2024-Swadge-FW/commit/2a320ffe53023b4d093c6d21bc041776fc6191fc). Before I merge your PR, could you check if that change also kinda works? I get the feeling that no matter where we land, its gonna break on someone's machine and we'll patch it later anyway.

Huh, yeah that pretty much does exactly what I described before and also in a much simpler way... Oh, I guess I tried that and it broke because I have an asset with a space in the filename and I didn't just rename it. So, yeah, not sure if the spaces problem has an easy fix but also not sure if we care.

So yeah this way works for me. I'll still put up the asset preprocessor PR once this gets merged but I'll leave out the makefile changes with the timestamp file stuff since it seems superfluous and overcomplicated now...

dylwhich (approved):

AEFeinstein:
Score one for fever dream! I'm doing a bit more makefile tidying because I feel compelled to any time I look at it. Can't wait for the preprocessor improvements too!

PR #446, opened 2025-05-15
Name lists for Users and Swadgesonas

johnnywycliffe:
### Description

Name generator to bypass the 'free text entry' problem. Also shrinks each name down to 4 bytes for easy sending over swadgepass.

### Test Instructions

`./swadge_emulator -m "NameTest"`

It is set to print a whole list of names auto-generated by the system, then the user can set the name to whatever they want.

There's two modes: regular and User:
- Regular gives you unlimited name setting. Used for Swadgesonas and the like
- User locks the user into `listLen >> 2` (1/4) options and locks the end code completely. 

Switching between these can be accomplished with "B" button, but this isn't something that will happen during normal usage. As such, I don't force a refresh. Save the new code with "A" immediately after pressing "B" to avoid unexpected behavior.

### Ticket Links

None.

### Readiness Checklist


main/utils/nameList.c
```diff
@@ -0,0 +1,296 @@
...
+// Defines
+//==============================================================================
+
+#define USER_LIST_SHIFT 1
+
+//==============================================================================
+// Consts
+//==============================================================================
+
+static const char* const adjList1[] = {
```
AEFeinstein, line 37: May want to run words by #super-accessibility to see if anyone would get offended.

main/utils/nameList.c
```diff
@@ -0,0 +1,296 @@
...
+{
+    _getMacAddress();
+    arrayLens[0] = ARRAY_SIZE(adjList1);
+    arrayLens[1] = ARRAY_SIZE(adjList2);
+    arrayLens[2] = ARRAY_SIZE(nounList);
+}
+
+void generateMACUsername(nameData_t* nd)
+{
+    nd->idxs[ADJ1] = baseMac[2 + ADJ1] % arrayLens[ADJ1]; // Functionally equivalent to _mutateIdx() w/ idx = 0
```
AEFeinstein, line 102: This uses indices 2, 3, and 4, but I think you want indices 3, 4, and 5. If you want to extract four, six bit fields from the bottom three bytes of a mac you can use this tricksy union. Write to `macAddr` and read from the `bits` fields.
```c
typedef union __attribute__((packed))
{
    struct
    {
        uint8_t macAddr[6];
    };
    struct __attribute__((packed))
    {
        uint8_t top1      : 8;
        uint8_t top2      : 8;
        uint8_t top3      : 8;
        uint8_t bits18_23 : 6;
        uint8_t bits12_17 : 6;
        uint8_t bits6_11  : 6;
        uint8_t bits0_5   : 6;
    };
} macBits_t;
```

main/swadge2024.c
```diff
@@ -406,6 +407,9 @@ void app_main(void)
         cSwadgeMode->fnEnterMode();
     }
 
+    // Initialize username settings
+    initUsernameSystem();
```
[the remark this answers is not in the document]
johnnywycliffe: It only needs to be called once per Swadge, stores MAC and length of the name arrays, then never needs to be called again. 

I figured there was more appropriate than saying everyone needs to call it just in case it wasn't set by someone else already

main/utils/nameList.c
AEFeinstein: Sure, its just a pattern to adjust to your need. Still, I'd avoid using bytes 1, 2, and 3.

main/CMakeLists.txt
```diff
@@ -48,6 +48,7 @@ idf_component_register(SRCS "asset_loaders/common/heatshrink_encoder.c"
                             "modes/test/factoryTest/factoryTest.c"
                             "modes/test/keebTest/keebTest.c"
                             "modes/test/touchTest/touchTest.c"
+                            "modes/test/.c"
```
AEFeinstein, line 51: A filename?

johnnywycliffe: I don't k ow what happened here tbh. I'll fix it before it's done.

AEFeinstein: Oh right, still in draft

main/utils/nameList.c
johnnywycliffe: I decided to just XOR everything, we still need four bytes

main/utils/nameList.c
```diff
@@ -0,0 +1,487 @@
...
+// Drawing functions
+
+static void _drawWordCenteredOnX(int xCenter, int yOffset, const char* str, paletteColor_t col)
+{
+    int xCoord = xCenter - (textWidth(&fnt, str) >> 1);
+    int yCoord = ((TFT_HEIGHT - fnt.height) >> 1) + yOffset;
+    drawText(&fnt, col, str, xCoord, yCoord);
+}
+
+static void _drawFadingWords(nameData_t* nd)
```
AEFeinstein, line 432: These colors are hot

AEFeinstein (approved):

main/utils/nameList.h
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * @brief Generates a username based on the MAC of the swadge
+ *
+ * @param nd Data struct the name is saved to
+ */
+void generateMACUsername(nameData_t* nd);
+
+/**
+ * @brief Generates a random username. If user is true, locks the end code.
```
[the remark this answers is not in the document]
AEFeinstein: Oh, it's a field in `nameData_t`

main/modes/test/nameTest/nameTest.c
```diff
@@ -0,0 +1,94 @@
...
+        if (evt.down && evt.button & PB_B)
+        {
+            nt->nd.user = !nt->nd.user;
+        }
+        finished = handleUsernamePickerInput(&evt, &nt->nd);
+    }
+
+    if (finished)
+    {
+        // do more code
```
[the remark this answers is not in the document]
johnnywycliffe: Nope! This is the example file. That's just a 'here's where you'd put code... if you had any'

main/utils/nameList.h
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * @brief Sets the username from a predefined nd
+ *
+ * @param nd nd containing the data
+ */
+void setUsernameFromND(nameData_t* nd);
+
+/**
+ * @brief Set the Username From indexs. Useful for loading data from swadgepass
```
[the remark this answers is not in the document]
johnnywycliffe: Yes it is. Will feeeeeex

main/utils/nameList.h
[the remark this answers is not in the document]
johnnywycliffe: And used to be a variable. Should prooooobably fix that.

main/modes/test/nameTest/nameTest.c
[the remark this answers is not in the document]
AEFeinstein: Alright there Dinkleberg

main/utils/nameList.c
AEFeinstein: I never thought to do interesting math on the palette values...

AEFeinstein (approved):

main/utils/nameList.c
johnnywycliffe: It was that or 36 draw calls.

PR #447, opened 2025-05-17
Asset preprocessor improvements

### LINKED ISSUE #400: Asset Preprocessor Improvement

dylwhich:
### Description

This is a fairly extensive overhaul of the assets preprocessor, slightly beyond what's described in #400.
Main changes:
* `assets.conf` -- now there's a config file! you don't need to modify the C code if you want to use an existing processor with a new file extension
* `.opts` files -- you can use `<filename>.opts` to change how `<filename>.<ext>` is processed, or `dir/.opts` to affect everything inside `dir/`
* Exec processors -- you can define an asset processor that just runs a shell command now, which I've set up for sokoban as an example. This should be great for cases where you can just use a python script.
* Assets are only processed as needed using file modification the same way make does
* All file opening is abstracted out of the asset processors themselves, so they can input and output using either file handles or preloaded buffers of file data.

### Test Instructions

Tested by converting all the existing asset processors of the new system.

Needs some testing on non-Linux but there shouldn't have been much platform-specific. I checked all the GitHub builds and all the asset images have the same size now so that's a good sign.

Read the docs and see if it makes any sense! I tried to split it between "just configuring a new file type" and "i need  to code up a new processor".

### Ticket Links

Closes #400 

### Readiness Checklist


AEFeinstein:
Gimme a few days to review this please

dylwhich:
> Gimme a few days to review this please

Take your time! It looks like the OSX version is somehow generating a `cnfs_image.c` that's over 4GB so I'll need to track that down in the meantime...

dylwhich:
Okay, got that sorted out! Forgot to set the output length on the `.bin` processor and GCC just had less obvious Undefined Behavior than clang did.

Side note, I'm annoyed a the slowness of the windows IDF build again and I came across this: https://www.reddit.com/r/esp32/comments/1eb874g/comment/leu17ii/ which links to https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/build-system.html#optional-project-variables which it doesn't look like we've tried... Maybe I'll see if I can trim that down a bit.

tools/assets_preprocessor/src/assets_preprocessor.h
```diff
@@ -0,0 +1,575 @@
...
+ * \subsection assetProc_structExecs Exec Preprocessors
+ *
+ * In addition to the function-based implementation used by most processors, they may also
+ * be defined as an processorType_t::EXEC processor, which executes some other program (e.g.
+ * a Python script) in order to process an asset. It's very important to keep in mind that
+ * this program must be available on all platforms in order for assets to process properly.
+ * If an executable that will be called by an exec processor must be compiled, it must be
+ * added to the makefile in order to ensure it is built before the asset processor runs.
+ *
+ * An EXEC processor is defined by a command string in its assetProcessor_t::exec field.
```
AEFeinstein, line 144: Same for this `assetProcessor_t::exec`

AEFeinstein (changes requested):
You're 95% of the way there, and the last 5% is mostly documentation weirdness. Actual code changes look good! Dithering option is my favorite feature here.

assets.conf
```diff
@@ -0,0 +1,57 @@
...
+
+[.font.png]
+inExt = font.png
+outExt = font
+func = font
+
+; Compressed JSON files
+[.json]
+outExt = json
+func = json
```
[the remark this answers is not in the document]
dylwhich: Technically it could have been but instead I ended up changing it so the JSON processor actually parses the file and an invalid JSON file fails the build! And along the way I finally tracked down a weird bug that happens with some small assets when the heatshrink data ends up larger than the input data.

tools/assets_preprocessor/src/assets_preprocessor.h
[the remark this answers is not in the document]
AEFeinstein: I'll trust you to sort this one out. Sorry doxygen is being a pain

tools/assets_preprocessor/README.md
```diff
@@ -6,15 +6,137 @@ Usage:
...
+| `%a`        | Input file extension, without leading '.'  |
+| `%b`        | Output file extension, without leading '.' |
+| `%%`        | Literal `%` character                      |
+
+## <a name="options-files">Options Files</a>
+
+Some asset processors support the use of options files, which can be used to configure
+how a specific asset or directory of assets are processed. To set the options for a single
+asset, such as `myImage.png`, create a text file with the same name as the asset file but
+with the extension `.opts`, such as `myImage.opts`. This file is formatted as an [INI file][ini],
```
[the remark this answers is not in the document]
dylwhich: Oops, that's supposed to be wikipedia! Forgot to add it to the references table...

dylwhich:
Ugh, the weird parsing differences between doxygen's handling of `.md` files and the inline documentation might be too much. And it's so awkward to link between them... I think I'm going to put pretty much everything into the doxygen for `assets_preprocessor.h` with separate "how to implement a processor" and "how to configure the processor" sections, and maybe just let the README document how we have it configured. Or just duplicate a little bit of the docs into the README...

AEFeinstein:
Yeah, I messed around a little with `README.md` too and this is the first time I feel like Doxygen is failing me. Why can't it just treat it like a normal file?

Wait, new feature to the rescue? Assuming you have Doxygen 1.13.2, you can run `doxygen -u Doxyfile` to update the file with new options, then try setting `IMPLICIT_DIR_DOCS` to no? I think that might make it a normal, linkable file, like `{#design_principles}`. Other new settings include
- `AUTOLINK_IGNORE_WORDS`
- `HIDE_UNDOC_NAMESPACES`
- `EXTERNAL_TOOL_PATH`
- `WARN_LAYOUT_FILE`
- `IMPLICIT_DIR_DOCS`
- `SHOW_ENUM_VALUES`
- `RTF_EXTRA_FILES`
- `PLANTUMLFILE_DIRS`

tools/assets_preprocessor/src/assets_preprocessor.h
dylwhich: Hmm, this one doesn't work either way. Doxygen _hates_ anonymous structs and unions, I don't know why it can't just treat them the same way C does. I don't think it's actually possible to generate a reference link to `assetProcessor_t::exec` or `assetProcessor_t::function` from doxygen. It's also not possible to put a literal percent inside a code block, \`\%\` gets you a backslash and percent, and \`%\` gets you an empty code block...

dylwhich: Oh wait, I guess the percent does work? Neat, not sure what I did though.

jarettmillard: You complained about it. That makes the problem go away on its own at least half the time.

dylwhich:
> Assuming you have Doxygen 1.13.2

Looks like that hasn't made it to Fedora yet :(

I ended up just duplicating most of what's in the README into `assets_preprocessor.h` and trying to separate it out better into usage vs. implementation sections. I'm not going to do anything with the README since I think it's still nice to have outside of doxygen... Also I think I fixed up all the remaining doxygen weirdness but I'm sure I probably missed something.

AEFeinstein:
> Looks like that hasn't made it to Fedora yet :(

Ubuntu neither, but that's why [the instructions and CI download it right from the source](https://github.com/AEFeinstein/Super-2024-Swadge-FW/blob/789331747310a1d71b76978c2d2e601183f7234c/.github/workflows/static.yml#L40)! Either way, I fiddled some more and it still doesn't work as I want it to. `IMPLICIT_DIR_DOCS` controls whether or not `README.md` is shown as directory documentation or not, but in the not case it isn't shown _at all_. I was expecting the file. Oh well. [Their own documentation](https://doxygen.nl/manual/markdown.html#md_page_header) has notes about including markdown readmes, but that doesn't work either.

tools/assets_preprocessor/src/assets_preprocessor.h
```diff
@@ -48,10 +36,239 @@
...
+ * \paragraph assetProc_chart chart
+ * Processes the input file as a Clone Hero chart, which can be created by
+ * <a href="https://efhiii.github.io/midi-ch/">this tool</a>. See also the
+ * <a
+ * href="https://github.com/TheNathannator/GuitarGame_ChartFormats/blob/main/doc/FileFormats/.chart/Core%20Infrastructure.md">
+ * .chart file spec</a>.
+ *
+ * \paragraph assetProc_font font
+ * Processes special font PNG files, which can be created by the
+ * <a href="https://github.com/AEFeinstein/Super-2024-Swadge-FW/blob/main/tools/font_maker/README.md">
```
AEFeinstein, line 156: We can include this `README.md` in doxygen, but if only we could link it sanely :(

AEFeinstein (approved):

AEFeinstein:
Well this one is on me. Because `cnfs_image.h` isn't checked in anymore, we get a broken link when building docs before source, which is what the GitHub Action does. D'oh. 
```
/home/adam/git/Super-2024-Swadge-FW/main/utils/cnfs.h:19: warning: unable to resolve reference to 'cnfsFileIdx_t' for \ref command
```
Maybe we should merge `build-firmware-and-emulator.yml` and `static.yml`, but I don't want to think about CI right now :sweat_smile:

tools/assets_preprocessor/src/assets_preprocessor.h
dylwhich: Leaving this here, there's a few options I don't think I quite tried before so maybe we can fix this after all...

https://stackoverflow.com/a/26710792

Doxyfile
```diff
@@ -954,7 +954,6 @@ INPUT                  = ./main \
                          ./main/asset_loaders/heatshrink_helper.h \
                          ./main/asset_loaders/heatshrink_helper.c \
                          ./tools/assets_preprocessor/src \
-                         ./tools/assets_preprocessor/README.md \
```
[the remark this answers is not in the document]
dylwhich: That works for me too, I just figured it was all duplicated in the doxygen so I wouldn't have multiple pages for the same thing. I think this is just going to add all those markdown files to the main navbar though? Not sure if there's a way to add them to like a sublist or something like that...

tools/assets_preprocessor/src/assets_preprocessor.h
AEFeinstein: You can try, but it feels the issue here is specific to it being a `README.md` not `RANDOM_NAME.md` and Doxygen treating it like directory do.

Wait, maybe this? https://stackoverflow.com/a/37716243

Issue #416, opened 2025-01-31
Utility - SwadgePass


emilyanthony4244:
## Summary
A utility for swadge users to pass their information to another swadge similar to StreetPass. Copypasta from Wikipedia:

> StreetPass was a [Nintendo 3DS](https://en.wikipedia.org/wiki/Nintendo_3DS) functionality which allows passive communication between Nintendo 3DS systems held by users in close proximity, an example being the sharing of [Mii](https://en.wikipedia.org/wiki/Mii) avatars in the [StreetPass Mii Plaza](https://en.wikipedia.org/wiki/StreetPass_Mii_Plaza) application, and other game data. Similar to SpotPass in the Nintendo 3DS, when new data is received from StreetPass and the system is in sleep mode, the notification LED of the system will turn green and will remain green until the system is taken out of sleep mode; when StreetPass data is received while the system is not in sleep mode, the notification LED will blink green a few times before returning to being off.
> 
> StreetPass allows users to exchange software content from select games played on their system, regardless of what software is currently in the console. StreetPass functionality must be activated for each piece of compatible software and can be disabled by parental controls. Currently shared content is stored in one of twelve "data slots" in the console. Using this data slot, Nintendo 3DS users can readily share and exchange content for multiple games at the same time, whenever they are connected.[[5]](https://en.wikipedia.org/wiki/SpotPass_and_StreetPass#cite_note-8) Using the console's background connectivity, a Nintendo 3DS in Sleep Mode can automatically discover other Nintendo 3DS systems within range, establish a connection, and exchange content for mutually played games, all transparently and without requiring any user input. For example, in [Rhythm Heaven Megamix](https://en.wikipedia.org/wiki/Rhythm_Heaven_Megamix), if the user passes by someone with the same software, they will take on a figure-fighting duel challenge. Each game can only hold a certain number of StreetPass exchanges, requiring the player to check their game software before additional exchanges can be made (for example, StreetPass Mii Plaza can only hold ten visitors at a time).[[6]](https://en.wikipedia.org/wiki/SpotPass_and_StreetPass#cite_note-9)

I am envisioning a similar SwadgePass system, hoping to capture user data (qty ????, Nintendo used 12, no idea what we would be capable of) as a collecting card type of system.

## Technical Spec
This has a few prerequisites to function correctly:

1. A common data structure of username, game information, trophies, high scores whatever that is stored in non volatile memory and shared between modes. This requires its own ticket.
[...]

PR #445, opened 2025-05-14
The Trophy System

johnnywycliffe:
### Description

Do you like Steam achievements? Playstation Trophies? Whatever Xbox calls their stuff?

Now the Swadge has that too!
- Four types of trophies: Single trigger, Additive, Progressive, and checklists
- Many options: You can hide them, give them a difficulty rating, add titles, descriptions...
- NVS friendly: Automatically handles all the NVS stuff so you don't have to think about it
- Small: Data needs to be constant, so it doesn't get into RAM, and this code enforces that
- Helpers: Checking and setting bit flags easier, Getting the points for one mode or several, getting the number of trophies...
- Gamer Score: Automatically creates a gamer score, each mode totaling up to 1000 points. Good for competition!
- Automatic list displays: Built-in displays for showing the trophies with minimal setup
- Final completion trophy: Automatically adds a completion trophy
- Documentation: I wrote a tutorial on how to set it up in doxy as part of trophy.h

This will also feed into swadgepass, both the score and the last won trophy are saved automatically and can be sent relatively cheaply to another swadge.

### Test Instructions

`./swadge_emulator -m "Trophy Test"`

Also, there's a "Trophy Case" mode on the main menu. Go ahead and try that out too.

### Ticket Links

#397 

### Readiness Checklist
- [x] I have run `make format` to format the changes
- [x] I have compiled the firmware and the changes have no warnings
- [x] I have compiled the emulator and the changes have no warnings
- [x] I have run `make cppcheck` and checked that `cppcheck_result.txt` has no warnings for the changes
- [x] I have added doxygen comments to any code used by more than one Swadge mode. This includes `/*! \file` comments with Design Philosophy, Usage, and Example sections for new headers.
- [x] I have run `make docs` and checked that `doxy_warnings.txt` has no warnings for the new code

PR #448, opened 2025-05-20
Swadgepass

AEFeinstein:
## Description

Added SwadgePass transmission and reception when in LED idle mode. Added the capability for modes to retrieve SwadgePass data. Added `fnAddToSwadgePassPacket` to `swadgeMode_t` to build a packet to send. This is all documented in `swadgePass.h` and docs can be built with `make docs`.

Refactor `dance.c` to split LED animations into separate files. The file was getting a bit long for my liking. Note that I took a single-file-header approach here, which means function source is in the header, but guarded by `DANCE_IMPLEMENTATION`. This means `DANCE_IMPLEMENTATION` **must only** be declared once, in `dance.c`, and the headers can be included anywhere. See https://github.com/nothings/stb/blob/master/docs/stb_howto.txt

## Test Instructions

1. Take two Swadges and leave them in LED idle mode next to each other
2. Wait about ten minutes, to be safe
3. Unlock the debug menu with the Konami Code on the main menu
4. Check in the SwadgePass Test mode that data has been received from the other Swadge
5. Select the data in the menu and then toggle the 'used' state with the A button
6. Restart the Swadge, re-enter the SwadgePass Test mode, and verify that the used state persists

## Ticket Links

- #416 

## Readiness Checklist


main/utils/swadgePass.h
```diff
@@ -0,0 +1,159 @@
...
+struct swadgeMode;
+
+//==============================================================================
+// Function Declarations
+//==============================================================================
+
+void fillSwadgePassPacket(swadgePassPacket_t* packet);
+void sendSwadgePass(swadgePassPacket_t* packet);
+void receiveSwadgePass(const esp_now_recv_info_t* esp_now_info, const uint8_t* data, uint8_t len, int8_t rssi);
+
```
johnnywycliffe, line 154: May need a method to clear NVS data if it gets too full/if we only want to save a certain number of passes at a time.

johnnywycliffe (commented):
I have two concerns/scenarios I'd like to bring up in regards to the current functionality of the swadges.

1. There's no duplicate rejection, If two people sit int he same room with their swadges on, it will continuously write and overwrite the data. Maybe the connection rate is too low for us to care? 
2. We're just going to store every user that does connect? Swadgepass packets aren't going to be large, I get that, but also there's going to be thousands of these at MAGFest. The DS only had a few saved at a time IIRC. It doesn't appear like anything clears previously read passes, or a way to allow a player to clear pass data they don't want for whatever reason.

.github/pull_request_template.md
```diff
@@ -1,19 +1,31 @@
...
 - [ ] I have added doxygen comments to any code used by more than one Swadge mode. This includes `/*! \file` comments with Design Philosophy, Usage, and Example sections for new headers.
 - [ ] I have run `make docs` and checked that `doxy_warnings.txt` has no warnings for the new code
+
+### Feature Usage
+
+<!--- These aren't mandatory, especially for non-game tickets, but are strongly encouraged -->
+
+- [ ] As much hardware input is used as reasonable (buttons, touchpad, IMU, microphone)
+- [ ] As much hardware output is used as reasonable (screen, LEDs, speaker)
+- [ ] [Trophy](https://adam.feinste.in/Super-2024-Swadge-FW/trophy_8h.html) support is integrated
```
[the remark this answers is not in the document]
AEFeinstein: It'll exist when #445 is merged. I'm getting ahead of the ball here.

main/modes/utilities/dance/dance.c
```diff
@@ -282,21 +312,53 @@ void danceMainLoop(int64_t elapsedUs)
...
+            danceState->maySleep = true;
+            // Turn off wifi before sleeping
+            espNowPreLightSleep();
+        }
+    }
+
+    danceState->bcastTimer -= elapsedUs;
+    if (danceState->bcastTimer <= 0)
+    {
+        // Broadcast every 7.5s, +/- 1.875s
```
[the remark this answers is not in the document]
AEFeinstein: I did a bunch of simulation of different delays and randomness and this had the best results while keeping the duty cycle low enough for power constraints. I have the python script somewhere, not the `tools` folder though.

main/utils/swadgePass.h
AEFeinstein: Hm, I don't want to give a single mode the easy ability to wipe data other modes might care about.

johnnywycliffe: We could tie it to the current swadgeMode_t? That's what I'm doing with the trophy system to avoid conflicts.

AEFeinstein:
> There's no duplicate rejection, If two people sit int he same room with their swadges on, it will continuously write and overwrite the data. Maybe the connection rate is too low for us to care?

The MAC address is used as the NVS key, so you'll never have two packets from the same Swadge. Otherwise overwriting old data is a feature, not a bug! There may be a new high score or avatar, or at least the "has this data been used" flags get cleared, just like re-tags in StreetPass (though there's a bit of a time delay there that doesn't exist here)

> We're just going to store every user that does connect? Swadgepass packets aren't going to be large, I get that, but also there's going to be thousands of these at MAGFest. The DS only had a few saved at a time IIRC. It doesn't appear like anything clears previously read passes, or a way to allow a player to clear pass data they don't want for whatever reason.

Yeah, good idea, I should add a cap. 100 passes seems like a number.

johnnywycliffe:
How long is it on average to get a pass if there's only two? If it's every ten minutes, then I guess it'll be fine. Something something write cycles something something we're never going to hit a limit like that.

If we have a cap of 100, can a user clear space for new ones? That seems like something we may want to do.

main/utils/swadgePass.h
johnnywycliffe: Nevermind, I got it confused. I understand, but the mode should be able to handle not having swadgepass data or it'll crash whenever there's no swadgepasses around it. Like in testing. It shouldn't ever delete data in use, obviously, but maybe for use in clearing old passes out?

johnnywycliffe:
Lastly, would a marker indication of which data is new may be helpful to indicate to users so the data deosn't get drowned in a sea of 100 passes looking for anything that's changed.

johnnywycliffe:
Lastly lastly, Just a note that we might want to override the first LED in line to indicate a new message in the future. Not sure how you'd want to implement that.

main/utils/swadgePass.h
AEFeinstein: When adding the cap, I'll have new passes evict the most used older packet. Good?

johnnywycliffe: Good enough for me.

AEFeinstein:
> How long is it on average to get a pass if there's only two? If it's every ten minutes, then I guess it'll be fine. Something something write cycles something something we're never going to hit a limit like that.

I think the simulation put it closer to a minute, but I haven't measured it in practice. I don't think we'll be burning out flash, especially because duplicates shouldn't overwrite (I'll have to tweak that).

> If we have a cap of 100, can a user clear space for new ones? That seems like something we may want to do.

I'm going to have newer data evict older data based on which are the most used by modes (most bits set). Maybe we'll need a 'favorite' feature to keep specific data around...

> Lastly, would a marker indication of which data is new may be helpful to indicate to users so the data doesn't get drowned in a sea of 100 passes looking for anything that's changed.

That's up to the individual modes to flag what's been used and either not display used data, or sort it to their liking.

> Lastly lastly, Just a note that we might want to override the first LED in line to indicate a new message in the future. Not sure how you'd want to implement that.

We'll see what LEDs we get, but in general I'm not jazzed about a SwadgePass LED. If the system is working, users should be getting new data with reasonable frequency and the LED would be on more often than not, which isn't too informative. If data is sparse, then the LED becomes more useful but the feature is less fun.

johnnywycliffe:
Comments read, in agreement, no notes.

main/modes/test/swadgePassTest/swadgePassTest.c
```diff
@@ -0,0 +1,208 @@
...
+
+    // Get SwadgePasses
+    getSwadgePasses(&spt->swadgePasses, &swadgePassTestMode, true);
+
+    // Initialize menu
+    spt->menu = initMenu(swadgePassTestName, swadgePassTestMenuCb);
+
+    // Add all keys to the menu
+    node_t* passNode = spt->swadgePasses.first;
+    while (passNode)
```
dylwhich, line 80: Not a huge issue for the test mode but if there are no swadge passes at all, the empty menu seems to cause a crash after a few seconds.

main/swadge2024.h
```diff
@@ -349,6 +353,17 @@ typedef struct
      * globalMidiPlayerFillBuffer() will be used instead to fill sample buffers
      */
     fnDacCallback_t fnDacCb;
+
+    /**
+     * @brief This function is called to fill in a SwadgePass packet with mode-specific data. The Swadge mode should
+     * only fill in it's relevant data and not touch other mode's data.
```
dylwhich, line 359: Is this just going to be one struct that every mode shares and adds its own members to as needed?

dylwhich (approved):

dylwhich:
I did a little bit of testing on a hotdog and protodog and everything seems to be working nice!y! I left them for maybe 5 minutes first and one had gotten a pass but the other hadn't. Then I left them for around one minute and checked and that swadge had already gotten the pass. Really neat!

I also checked in the emulator and that indeed works too. It is a little weird with the `nvs.json` so maybe I should bring back the argument to use a different file to make emu swadgepass testing easier...

main/modes/test/swadgePassTest/swadgePassTest.c
AEFeinstein: Hrm, I'm not replicating that with the emulator (deleted `nvs.json`) or firmware (did a factory reset from the secret menu). Maybe you had stale data or something?

main/swadge2024.h
AEFeinstein: Yep! The alternative would be giving letting each mode append to a pointer before transmission, then letting each mode strip off bytes from a pointer on reception. That feels way more error prone though, since it requires symmetry in the write & read functions for each mode.

AEFeinstein:
Hm, I don't follow why `nvs.json` weird. For testing I'd shim this (or similar) in `app_main()` before the loop:
```c
initSwadgePassReceiver();
uint8_t dummyMac[]       = {1, 2, 3, 4, 5, 6};
const esp_now_recv_info_t enri = {
    .des_addr = dummyMac,
    .src_addr = dummyMac
};
swadgePassPacket_t dummyPacket = {
    .preamble = 0x5350,
    .version  = 0,
};
receiveSwadgePass(&enri, (const uint8_t*)&dummyPacket, sizeof(dummyPacket), 0);
deinitSwadgePassReceiver();
```

johnnywycliffe (approved):
Looks good.

PR #452, opened 2025-05-24
Added usernamer to swadgePass

johnnywycliffe:
## Description

Adds the username to the swadgepass struct

## Test Instructions

Verified saved username was sent to another swadge in the emulator (checked nvs.json)

## Ticket Links

None

### Code Quality

### DIFF
--- a/main/utils/nameList.c
+++ b/main/utils/nameList.c
@@ -152,8 +152,8 @@ void initUsernameSystem()
 
     // Initialize
     swadgeUsername.user = true;
-    size_t blobLen = 3 * sizeof(int8_t);
-    if(!readNvsBlob(nvsKeys, &swadgeUsername.idxs, &blobLen))
+    size_t blobLen      = 3 * sizeof(int8_t);
+    if (!readNvsBlob(nvsKeys, &swadgeUsername.idxs, &blobLen))
     {
         generateMACUsername(&swadgeUsername);
         writeNvsBlob(nvsKeys, swadgeUsername.idxs, 3 * sizeof(int8_t));
--- a/main/utils/nameList.c
+++ b/main/utils/nameList.c
@@ -67,6 +67,8 @@ static const char* const nounList[] = {
     "woodwind", "drummer",  "rocker",   "wolf",    "snake",    "driver",  "seahorse", "artist", "foxbat",  "eagle",
 };
 
+static const char nvsKeys[] = "username";
+
 //==============================================================================
 // Enums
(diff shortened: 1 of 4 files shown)


johnnywycliffe:
I call this one "More work for Adam just because!"

main/utils/nameList.c
```diff
@@ -146,6 +149,16 @@ void initUsernameSystem()
     listLen[1] = ARRAY_SIZE(adjList2);
     listLen[2] = ARRAY_SIZE(nounList);
     _getMacAddress();
+
+    // Initialize
+    swadgeUsername.user = true;
+    size_t blobLen      = 3 * sizeof(int8_t);
```
AEFeinstein, line 155: For one, this'd be safer as `sizeof(swadgeUsername.idxs)`, since that's what you're actually writing into. But also if you're only saving three bytes, you may as well use the 32 bit NVS functions. [There's some extra overhead for variable length NVS entries](https://docs.espressif.com/projects/esp-idf/en/v5.2.5/esp32s2/api-reference/storage/nvs_flash.html#structure-of-entry).

main/utils/swadgePass.h
```diff
@@ -121,6 +121,10 @@ typedef struct __attribute__((packed)) swadgePassPacket
     uint16_t preamble; ///< Two bytes that specifically begin a SwadgePass packet
     uint8_t version;   ///< A version byte to differentiate packets per-year
     struct
+    {
+        uint8_t nameIdxs[4];
```
AEFeinstein, line 125: Wait, are the indices three or four bytes?

AEFeinstein: Three indices and a MAC byte? That's a 32 bit number in a trenchcoat...

AEFeinstein:
> I call this one "More work for Adam just because!"

I don't get paid enough for this.

main/utils/swadgePass.h
johnnywycliffe: I should've just bitpacked. Blegh. I'll fix this in a minute.

johnnywycliffe:
Me neither.

johnnywycliffe:
Note if you're going to test this: Since the emulator changes it's MAC every time it reboots, you can't just compare the number saved to the NVS to the number in the packet. If you put a breakpoint after the packet is generated, uncomment the ESP_LOGI in nameList.c and compare, you can see that the first three numbers will be identical and the final number has changed to the current MAC address.

PR #473, opened 2025-07-29
Fix win ci

AEFeinstein:
## Description

Makefile tweaks to work around Windows shell line length limits

## Test Instructions

* Build with CI
* Build repeatedly and make sure it doesn't do a full build each time

## Ticket Links

n/a

### Code Quality

### Feature Usage

- [ ] As much hardware input is used as reasonable (buttons, touchpad, IMU, microphone)

### DIFF
--- a/makefile
+++ b/makefile
@@ -245,7 +245,6 @@ DEFINES = $(patsubst %, -D%, $(DEFINES_LIST))
 # Files to write compiler arguments to (workaround for Windows line limits)
 ################################################################################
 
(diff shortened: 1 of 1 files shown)


AEFeinstein:
@cnlohr I think this should do the trick. I was able to drop all the non-include path args into files, because those had a clean dependency on the makefile itself. I wasn't able to figure out a way for the makefile to only rewrite a file with the include paths only when the paths change, which meant it was rebuilding the entire program each time.

Issue #501, opened 2025-10-11
Swadgepass does not send username on hardware


johnnywycliffe:
## Description
After some testing with several swadges IRL, all swadges are sending `0` in the username field. In the emulator, the field was set and can be verified, but when the packet is send via ESPNow it fails. The packet is being sent, high scores are being transmitted, but the name field is always all zeroes.

## Expected behavior
Packed username is sent to the other device.

## Reproduction Steps
Load two swadges and have them send a swadgepass pakcet, then examine,

## Affected Version
 - Platform: Firmware
 - git branch: main
 - git tag: v3.0.0

AEFeinstein:
PR is #511 and has some feedback

PR #511, opened 2025-10-13
Namelist Fix

johnnywycliffe:
## Description

The username isn't properly set on current devices. To fix this, they need to be factory reset with a few tweaks.

Also, nameList could go out of bound when making the name for the user. Fixed that bug.

Note that this is a draft. The current fix is hacky.

## Test Instructions

Factory reset the swadge hardware, then test the nameList mode to see the options for that physical hardware.

## Ticket Links

#501 

### Code Quality

### DIFF
--- a/main/utils/nameList.c
+++ b/main/utils/nameList.c
@@ -381,16 +381,15 @@ void drawUsernamePicker(nameData_t* nd)
 
 nameData_t* getSystemUsername(void)
 {
-    int32_t packed = 0;
-    readNvs32(nvsKeys, &packed);
-    setUsernameFrom32(&swadgeUsername, packed);
     return &swadgeUsername;
 }
 
 void setSystemUsername(nameData_t* nd)
 {
     nameData_t data = *nd;
-    writeNvs32(nvsKeys, GET_PACKED_USERNAME(data));
+    int32_t packed = GET_PACKED_USERNAME(data);
+    writeNvs32(nvsKeys, packed);
+    setUsernameFrom32(&swadgeUsername, packed);
 }
 
 void setUsernameFrom32(nameData_t* nd, int32_t packed)


main/utils/nameList.c
```diff
@@ -400,7 +406,7 @@ void setUsernameFrom32(nameData_t* nd, int32_t packed)
 // Static Functions
 //==============================================================================
 
-static void _getMacAddress()
+static bool _getMacAddress()
```
AEFeinstein: This return value isn't being checked in `initUsernameSystem()`. Should it be before writing the username to NVS?

main/utils/nameList.c
```diff
@@ -378,6 +381,9 @@ void drawUsernamePicker(nameData_t* nd)
 
 nameData_t* getSystemUsername(void)
 {
+    int32_t packed = 0;
+    readNvs32(nvsKeys, &packed);
+    setUsernameFrom32(&swadgeUsername, packed);
```
AEFeinstein: Reading NVS each time this is called feels... wrong? Like `cosCrunchMainLoop()` calls this each loop when displaying high scores, which is a lot of reads. Is the goal really to call these functions each time, or to call them after some condition?

main/utils/nameList.c
johnnywycliffe: This is effectively for debug only. There's not really a reason to revert it since it could come in handy in the future.

main/utils/nameList.c
johnnywycliffe: The other option is more or less to assume it got initialized the first time. I have a deep suspicion of things that are supposed to be pre-set.

That said, I can def add an "initialized" bool to that checks if the name has been stored. It's just more memory.

main/modes/test/nameTest/nameTest.c
```diff
@@ -32,6 +34,12 @@ static void ntEnterMode(void)
     nt          = heap_caps_calloc(sizeof(ntData_t), 1, MALLOC_CAP_8BIT);
     nt->nd.user = false;
 
+    // FIXME: Crashes if not called again here
+    initUsernameSystem();
```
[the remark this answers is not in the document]
johnnywycliffe: The mode crashes if I don't. This is a double initialization, and I ned to see if I just need to initialize twice in general, or if it's a mode teardown thing.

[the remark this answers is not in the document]
johnnywycliffe: Answer: No. I fixed it somewhere along the line.

main/utils/nameList.c
johnnywycliffe: So, I wasn't saving the data to anywhere but the NVS. Updating the seave func to actually save to the struct fixed it.

I am a moron.

AEFeinstein (approved):

PR #523, opened 2025-10-16
Mega Pulse EX - vertical slice

JVeg199X:

### DIFF
--- a/.github/workflows/build-firmware-and-emulator.yml
+++ b/.github/workflows/build-firmware-and-emulator.yml
@@ -44,7 +44,7 @@ jobs:
               - nvs.json
             idf_install: ~/.espressif
           - os: osx-intel
-            runner: macos-13
+            runner: macos-15-intel
             family: osx
             emulator: true
             firmware: false


main/modes/games/platformer/mgEntity.c
```diff
@@ -3022,6 +3023,7 @@ void mg_updateBossDoor(mgEntity_t* self)
                 self->entityManager->playerEntity->xspeed = 32;
                 self->entityManager->playerEntity->state  = MG_PL_ST_NORMAL;
             }
+            break;
```
AEFeinstein: @JVeg199X I fixed the warnings in this commit. Adding this `break` was the only functional change. I wasn't sure if it was supposed to be here or not, but it seemed like it should be? If I was wrong, please delete it and add a "fallthrough" comment so the compiler doesn't complain

https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough

AEFeinstein (approved):

PR #532, opened 2025-10-26
Add factory test Eye LED pattern

AEFeinstein:
## Description

* Added factory test pattern for eye LEDs
* Add and use EYE_LED_W and EYE_LED_H
* Formatting

## Test Instructions

Run factory test mode and validate LEDs on hardware

## Ticket Links

n/a

### Code Quality

### Feature Usage

- [x] As much hardware input is used as reasonable (buttons, touchpad, IMU, microphone)

### DIFF
--- a/assets/ch32v003/matrix_factory.cfun
+++ b/assets/ch32v003/matrix_factory.cfun
@@ -11,7 +11,7 @@ typedef enum
 } factoryLedState_t;
 
 /**
- * @brief Main loop for ch32v003. Eyes blink and pupils move around periodically, with some randomness
+ * @brief Main loop for ch32v003. LEDs start all on, then sweep across rows, then sweep across columns
  *
  * @return 0
  */
--- a/main/modes/test/factoryTest/factoryTest.c
+++ b/main/modes/test/factoryTest/factoryTest.c
@@ -204,7 +204,7 @@ void testEnterMode(void)
     accelSetRegistersAndReset();
     accelPerformCal();
 
-    // Load some default firmware that blinks eyes
+    // Load firmware that turns all LEDs on, then sweeps horizontal and vertical test patterns
     ch32v003RunBinaryAsset(MATRIX_FACTORY_CFUN_BIN);
 
     // Set NVM to indicate test not passed yet


assets/ch32v003/matrix_factory.cfun
```diff
@@ -0,0 +1,106 @@
...
+
+typedef enum
+{
+    ALL_ON,
+    ROW_SWEEP,
+    COL_SWEEP
+} factoryLedState_t;
+
+/**
+ * @brief Main loop for ch32v003. Eyes blink and pupils move around periodically, with some randomness
```
johnnywycliffe: Not really important, but this description is wrong.

main/utils/nameList.c
```diff
@@ -387,7 +387,7 @@ nameData_t* getSystemUsername(void)
 void setSystemUsername(nameData_t* nd)
 {
     nameData_t data = *nd;
-    int32_t packed = GET_PACKED_USERNAME(data);
+    int32_t packed  = GET_PACKED_USERNAME(data);
```
johnnywycliffe: Ah yes, a single space.

assets/ch32v003/matrix_factory.cfun
```diff
@@ -0,0 +1,106 @@
...
+                    }
+                    break;
+                }
+                case COL_SWEEP:
+                {
+                    colsVisible++;
+                    if (colsVisible == EYE_LED_W + 1)
+                    {
+                        state      = ALL_ON;
+                        allOnTimer = 0;
```
johnnywycliffe: Since this immediately jumps to all on, it kinda seems like it gets stuck? Just because the solid mode doesn't have any difference from the last frame of the COL sweep.

Probably unnecessary, but maybe toggle all the lights off for half a second before putting them all on?

main/utils/nameList.c
jarettmillard: ` `

assets/ch32v003/matrix_factory.cfun
AEFeinstein: I see what you're saying, but the pattern serves the test purpose, checking that they all work, so this doesn't bother me. Realistically the factory only needs to check the all-solid pattern to validate hardware

Issue #557, opened 2025-11-27
Swadgesona Atrium


AEFeinstein:
## Summary

View Swadgesonas or something (placeholder)

AEFeinstein:
I kinda ignored your list but made a number of tweaks wrt. arrows, animations, and button inputs. It _feels_ better (at least to me), and I may have inadvertently fixed some of your list?

emilyanthony4244:
Yes - I'll have one more PR for you today ish.

PR #583, opened 2025-12-17
Atrium

emilyanthony4244:
## Description

Added atrium mode

## Ticket Links

#557 

### Code Quality

### Feature Usage

- [ ] As much hardware input is used as reasonable (buttons, touchpad, IMU, microphone)

### DIFF
--- a/main/modes/games/atrium/atrium.c
+++ b/main/modes/games/atrium/atrium.c
@@ -18,7 +18,7 @@
 #define MAX_SWADGESONA_IDXS (MAX_NUM_SWADGE_PASSES / SONA_PER) - (MAX_NUM_SWADGE_PASSES % SONA_PER) / SONA_PER
 #define ANIM_TIMER_MS       16667
 #define LOBBY_ARROW_Y       200
-#define LOBBY_BORDER_X      20
+// #define LOBBY_BORDER_X      20
 
 // Profile
 #define CARDTEXTPAD 4
@@ -118,7 +118,7 @@ static const char* const preambles[] = {
 };
 
 static const char* const editPromptText[] = {
-    "Choose Card", "Choose Identity", "Choose Location", "Pick Sandwich", "Save Profile",
+    "Choose Card", "Choose Identity", "Choose Location", "Pick Sandwich", "Save Profile", "Saved!",
 };
 
 // // Coordinates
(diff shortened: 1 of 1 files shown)


main/modes/games/atrium/atrium.c
```diff
@@ -0,0 +1,1292 @@
...
+                    else if (evt.button & PB_DOWN)
+                    {
+                        atr->down = false;
+                        if (atr->lbState < 2)
+                        {
+                            atr->lbState++;
+                        }
+
+                        atr->loadAnims = 0;
+                    }
```
AEFeinstein: This doesn't pass the sniff test. It's an `if`, not an `else if` without a line break between, and `PB_UP` and `PB_DOWN` are checked in both blocks, so `atr->up` and `atr->down` are both set and then immediately overwritten.

What was the intent here? The effect is equivalent to deleting the prior button checks and turning this into one large `if`/`else if` statement.

main/modes/games/atrium/atrium.c
```diff
@@ -0,0 +1,1292 @@
...
+
+// Draw
+static void drawArcade(uint64_t elapsedUs)
+{
+    atr->animTimer += elapsedUs;
+    if (atr->animTimer >= ANIM_TIMER_MS && atr->loadAnims < 20)
+    {
+        atr->animTimer = 0;
+        atr->loadAnims++;
+    }
```
AEFeinstein: This is fine as-is, and I'll point out for the future that we have a neat little timer macro
```c
/**
 * @brief Run timer_code every period, using tracking it with timer
 *
 * @param timer The accumulator variable, must persist between calls
 * @param period The period at which timer_code should be run
 * @param elapsed The time elapsed since this was last called
 * @param timer_code The code to execute every period
 */
#define RUN_TIMER_EVERY(timer, period, elapsed, timer_code) \
    do                                                      \
    {                                                       \
        timer += (elapsed);                                 \
        while (timer > (period))                            \
        {                                                   \
            timer -= (period);                              \
            {                                               \
                timer_code                                  \
            }                                               \
        }                                                   \
    } while (0)
```
And a simple example:
```c
// Read battery every 10s
RUN_TIMER_EVERY(menu->batteryReadTimer, 10000000, elapsedUs, { menu->batteryLevel = readBattmon(); });
```

PR #606, opened 2026-01-02
Mega Pulse EX final(???) tweaks

JVeg199X:
## Description

- Change word in Picross fact based on feedback.
- Fill in character art.
- Correctly dither credits image into websafe.
- Adjust level timers based on feedback.

- Cutscene dialogue adjustments
- removed 1cc requirement from time trophy and reset the saved timer when starting New Game +
- reduced thresholds for scoring trophies: now 400k and 1mil
- fix scroll locks not triggering at level start, only when player scrolls the screen in those directions.
- fix double Sure You Can. Sure You Can functions like a recovery move in Smash; once you use it, you can't double jump and MUST touch the ground before being allowed to do either move again.
- fixed bullet damage not resetting the combo multiplier

### DIFF
--- a/main/modes/games/platformer/mgCutscenes.c
+++ b/main/modes/games/platformer/mgCutscenes.c
@@ -154,6 +154,7 @@ static void setSongPitchesFromCurrentSong(mgGameData_t* gameData)
         }
         case MG_BGM_INTRO_STAGE:
         case MG_BGM_MAXIMUM_HYPE_CREDITS:
+        case MG_BGM_CLIMAX:
         {
             int16_t songPitches[]
                 = {62, 61, 60, 69, 62, 60, -1, -1}; // have 62 and 60 duplicated for some statistical weighting.


main/menu/menuMegaRenderer.c
```diff
@@ -428,10 +424,8 @@ void drawMenuMega(menu_t* menu, menuMegaRenderer_t* renderer, int64_t elapsedUs)
         {
             renderer->bgColorIdx = renderer->numBgColors - 1;
         }
-        if (renderer->conveyorBeltStyle)
-        {
-            renderer->yOff += renderer->bgColorIdx;
-        }
+        // scrolls the background
+        renderer->yOff += renderer->bgColorIdx;
```
AEFeinstein: @JVeg199X  please revert the menu changes. I don't want every menu background everywhere to scroll by default. This was intentionally optional.

AEFeinstein (changes requested):
Reviewed up to here, even though its still in draft.

DebrisHauler:
I'm working through a couple bugs again. Almost there.

DebrisHauler:
Noting something from @JVeg199X's work in this PR makes one turret in Ember's stage not fire. But it is fitting for the dump theme. And wasps still dive through the crumble blocks, but we think that's helpful anyway for Grind Pangolin's stage.

We're ready to ship it!

AEFeinstein:
@JVeg199X can you mark this ready for review if it is? And since I approved the draft  you should be able to squash and merge.

PR #664, opened 2026-09-10
Esp32s3 support

AEFeinstein:
## Description

- Retarget build for ESP32-S3
- Map GPIOs for all hardware platforms
- Update initialization for Fairy Prototype
- Better handle optional GPIOs, like `GPIO_VMON`
- Remove `CONFIG_SOUND_OUTPUT_BUZZER`
- Disable DAC when `SOC_DAC_SUPPORTED` isn't true. Left places for PDM implementation
- Adjust continuous ADC data type for S3
- Update generated docs for ESP32-S3
- Refactor `Kconfig.projbuild` files which configure build options (`idf.py menuconfig`) with dependencies and conditional defaults
- Added config for `CONFIG_ST7789_320x240` (unused)
- Autoformat Gotta Go

## Test Instructions

Build for S2 and S3. To flip between the two, run:

```bash
idf.py set-target esp32s2
```
or
```bash
idf.py set-target esp32s3
```

Then run

```bash
idf.py menuconfig
```

And select the hardware platform. Note that the hardware platform is dependent on the chip target.

## Ticket Links

n/a

### Code Quality

    - This is a lie, there are intentional warnings for now
    - This is a lie, there are intentional warnings for now

### Feature Usage

- [x] As much hardware input is used as reasonable (buttons, touchpad, IMU, microphone)
- [x] As much hardware output is used as reasonable (screen, LEDs, speaker)


main/swadge.c
```diff
@@ -679,27 +673,22 @@ static void initOptionalPeripherals(void)
...
     {
         setDacShutdown(false);
 
         // Otherwise initialize the battery monitor as a oneshot ADC
-        initBattmon(GPIO_NUM_6);
+        initBattmon(GPIO_VMON);
 
         // Initialize sound output if there is no input
-#if defined(CONFIG_SOUND_OUTPUT_SPEAKER)
         // Initialize the speaker. The DAC uses the same DMA controller for continuous output,
```
emilyanthony4244, line 687: I know the S3 has DMA controller upgrades that may enable us to use mic and speaker at the same time in the future. The new “General DMA” (GDMA) allows somewhat arbitrary mapping between channels and peripherals, whereas prior ESP32’s tended toward peripheral-specific DMA instances and channels (per Adafruit).

The ten peripherals on ESP32-S3 with DMA feature are SPI2, SPI3, UHCI0, I2S0, I2S1, LCD/CAM, AES, SHA,
ADC, and RMT, and there are 5 Tx/Rx channels for this.

For now I'm not proposing doing both at the same time, but here perhaps we should clarify in this comment that the **S2** requires the same DMA controller for continuous output and cannot support mic and speaker at the same time.

Ref for later:
https://documentation.espressif.com/esp32-s3_technical_reference_manual_en.pdf#dma

makefile
```diff
@@ -254,16 +254,16 @@ DEFINES_LIST = \
 	CONFIG_GC9307_240x280=y \
 	CONFIG_TFT_MAX_BRIGHTNESS=200 \
 	CONFIG_TFT_MIN_BRIGHTNESS=10 \
-	CONFIG_NUM_LEDS=6 \
+	CONFIG_NUM_LEDS=14 \
```
jarettmillard, line 257: Is this still needed if you added the defaults based on the hardware type?

jarettmillard (approved):

makefile
AEFeinstein: Yes. The `makefile` is for the emulator and `Kconfig.projbuild` is for the firmware.

emilyanthony4244 (commented):
Looks good to me, just tagged something we should think about and possibly clarify.

main/swadge.c
AEFeinstein: Good call out. I don't even know if @cnlohr's PDM solution will use DMA at all, so this may be moot anyway. It and/or MIDI synthesis will likely run on the S3's second core as well. We'll address restrictions when all that's in place.

